PR #1308, opened 2024-12-05
fix: load DAB drafts correctly

### LINKED ISSUE #1286: Loading DAB transactions needs to deserialize the certificate bytes back to string
When saving a transaction as a draft, the certificates are serialized to bytes in order to be added to the transaction. When loading from a draft, these certificates need to be deserialized in order to properly display the string to the user. 

For the Gossip CA Certificate, this is a simple reverse of the serialization process. 

For the GRPC Certificate, this will require extra metadata as the certificate itself is not stored in the transaction. The hash of the bytes of the string of the certificate is stored in the transaction, and this is not reversible. If there is an option to store extra metadata for a transaction, this would be the only way to save the text of the certificate. If not, then the user will see the hash, but not the certificate itself. If it isn't currently possible to display the GRPC certificate after loading from draft, then the user should still be able to easily see that the hash still exists.

Sheng-Long:
Signed-off-by: Tim Schmidt <tim@launchbadge.com>

**Description**:
Load DAB drafts correctly

**Related issue(s)**:

Fixes #1286 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/front-end/src/main/modules/ipcHandlers/utils.ts
+++ b/front-end/src/main/modules/ipcHandlers/utils.ts
@@ -6,6 +6,8 @@ import * as bcrypt from 'bcrypt';
 
 import { createHash, X509Certificate } from 'crypto';
 
+import { pki } from 'node-forge';
+
 import { getNumberArrayFromString } from '@main/utils';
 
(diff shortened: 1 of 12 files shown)


Sheng-Long:
Node update not loading correctly yet

PR #1327, opened 2024-12-10
feat: password inputs now show the 'reveal input' button by default.

### LINKED ISSUE #1291: When users are entering a password, allow them to view the characters they are entering
### Problem

Users were putting in the wrong password when moving to the new application for the recovery phrase. Users cannot view what they entering in the field currently. 

### Solution

Add an option to make the password entry visible to the user, if selected. You can see this commonly used in other apps. 

### Alternatives

_No response_

jbair06:
**Description**:
Password inputs have a reveal button added by default. In only a couple of cases, where a confirm password input is present, the reveal button is not visible. 

**Related issue(s)**:

Fixes #1291

### DIFF
--- a/front-end/src/renderer/components/ui/AppPasswordInput.vue
+++ b/front-end/src/renderer/components/ui/AppPasswordInput.vue
@@ -1,0 +1,61 @@
+<script setup lang="ts">
+import { computed, ref, useAttrs } from 'vue';
+
+import AppInput from './AppInput.vue';
+
+/* Props */
+const props = withDefaults(
+  defineProps<{
+    modelValue?: string | number;
+    filled?: boolean;
+    size?: 'small' | 'large' | undefined;
+    autoTrim?: boolean;
+    showIcon?: boolean;
+  }>(),
+  {
+    showIcon: true,
+  },
+);
+
+/* Emits */
+const emit = defineEmits(['update:modelValue']);
+
+/* Composables */
+const attrs = useAttrs();
+
+/* State */
+const isPasswordVisible = ref(false);
+
+/* Computed */
+const mergedProps = computed(() => ({
+  ...props,
+  ...attrs,
+}));
+
+/* Functions */
+const togglePasswordVisibility = () => {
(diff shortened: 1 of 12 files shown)


front-end/src/renderer/components/ui/AppPasswordInput.vue
```diff
@@ -0,0 +1,61 @@
...
+    autoTrim?: boolean;
+    showIcon?: boolean;
+  }>(),
+  {
+    showIcon: true,
+  },
+);
+
+/* Emits */
+const emit = defineEmits(['update:modelValue']);
```
SvetBorislavov, line 21: Emit is not used, clear it or use it

front-end/src/renderer/pages/OrganizationLogin/OrganizationLogin.vue
```diff
@@ -17,6 +17,7 @@ import { assertUserLoggedIn, getErrorMessage, isLoggedOutOrganization } from '@r
 import AppButton from '@renderer/components/ui/AppButton.vue';
 import AppInput from '@renderer/components/ui/AppInput.vue';
 import ForgotPasswordModal from '@renderer/components/ForgotPasswordModal.vue';
+import AppPasswordInput from '@renderer/components/ui/AppPasswordInput.vue';
```
SvetBorislavov, line 20: Swap the position of the ForgotPasswordModal import with this

SvetBorislavov (changes requested):
The automation tests are failing. It is probably because of the data-testids.
Could you ensure they exist after the change and are NOT DUPLICATED?
When binding the `attrs` to another element of the component (in your case in AppPasswordInput) to the AppInput, the properties may be duplicated to the root component's element and to the component that are passed

front-end/src/renderer/components/ui/AppPasswordInput.vue
jbair06: Is it not used on line 49?

jbair06:
Too many issues with rebasing, history is broken.

PR #1328, opened 2024-12-10
fix: updated color of the dropdown to be the same as other components

### LINKED ISSUE #1321: Import from recovery phrase in light mode is white
In the Settings Keys tab, the Import From Recovery Phrase tab uses white text, even in light mode. It is not visible. 

The text needs to match the other texts in light mode.

jbair06:
**Description**:
AppDropDown wasn't using the text-body class. This caused the text color to be incorrect in light mode.

**Related issue(s)**:

Fixes #1321 

**Notes for reviewer**:
Thanks for showing me how to do this, Svet!


RAMTO:
Screenshot 2024-12-10 at 11 07 55
I believe the text should be white here

SvetBorislavov (approved):

Issue #976, opened 2024-10-01
Error when duplicating tx's in a group and attempting to execute


yiliev0:
### Description

When duplicating a "Create Account" transaction multiple times(3 or more) within a transaction group and attempting to sign & execute, some transactions go through, but at some point, an error occurs with a toast message stating "Failed to add groupItem." The console error points to a unique constraint violation:
```Error: Error invoking remote method 'transactionGroups:addGroupItem': PrismaClientKnownRequestError: 
Invalid prisma2.groupItem.create() invocation in
/Users/yordaniliev/Desktop/hedera/3/hedera-transaction-tool/front-end/dist-electron/main/index.js:1604:34

  1601 }
  1602 const addGroupItem = async (groupItem) => {
  1603   const prisma2 = getPrismaClient();
→ 1604   return await prisma2.groupItem.create(
Unique constraint failed on the fields: (transaction_group_id,seq)```

### Steps to reproduce

1. Navigate to the Transaction Group page and create a new transaction group.
2. Add a "Create Account" transaction to the group.
3. Click "Duplicate" on the "Create Account" transaction 2 times or more
4. Attempt to sign & execute the transaction group.

### Additional context

**Actual Result:**
Some transactions execute successfully, but at a certain point, an error toast appears: "Failed to add groupItem."

**Expected Result:**
All transactions should be successfully signed and executed without triggering the unique constraint violation error.

### Hedera network

other

### Version

0.3.8

### Operating system

None

yiliev0:
Still reproducible

Issue #975, opened 2024-10-01
Error when saving empty group tx


yiliev0:
### Description

When attempting to save a transaction group with no transactions inside, an error toast message is displayed stating `Cannot read properties of undefined (reading 'groupId').`

### Steps to reproduce

1. Navigate to the Transaction Group page.
2. Create a new transaction group without adding any transactions and transaction group description.
3. Attempt to save the transaction group.

### Additional context

**Actual Result:**
An error toast message appears: "Cannot read properties of undefined (reading 'groupId')."

**Expected Result:**
Either a proper validation message should be displayed to inform the user that a transaction group cannot be saved without transactions, **or** the transaction group should be saved correctly even if no transactions are added.

### Hedera network

other

### Version

0.3.8

### Operating system

None

yiliev0:
Still reproducible

Issue #972, opened 2024-10-01
Incorrect navigation after discarding changes from group tx page


yiliev0:
### Description

When attempting to navigate to the Settings page from the Transaction Group page with unsaved changes, after clicking on "Discard" in the modal, the user is incorrectly navigated to the Transaction page instead of the Settings page.

### Steps to reproduce

1. Go to the Transaction Group page.
2. Make some changes without saving.
3. Attempt to navigate to the Settings page.
4. When prompted with the discard changes modal, click on "Discard."

### Additional context

**Actual Result:**
The user is navigated to the Transaction page after clicking "Discard."

**Expected Result:**
The user should be navigated to the Settings page after clicking "Discard."

### Hedera network

other

### Version

0.3.8

### Operating system

None

yiliev0:
Still reproducible

PR #1385, opened 2024-12-18
fix: transaction group bugs

SvetBorislavov:
**Description**:
This pull request fixes the following transaction group bugs:
- incorrect navigation when discarding a transaction group
- incorrect navigation when saving a transaction group
- error when saving empty transaction group
- inability to execute multiple duplicated transactions

**Related issue(s)**:

Fixes #
#972 
#975 
#976

### DIFF
--- a/front-end/src/renderer/stores/storeTransactionGroup.ts
+++ b/front-end/src/renderer/stores/storeTransactionGroup.ts
@@ -47,11 +47,13 @@ const useTransactionGroupStore = defineStore('transactionGroup', () => {
     description.value = group.description;
     const items = await getGroupItems(id);
     const drafts = await getDrafts(findArgs);
+    const groupItemsToAdd: GroupItem[] = [];
+
     for (const item of items) {
       const draft = drafts.find(draft => draft.id == item.transaction_draft_id);
       if (draft?.transactionBytes) {
         const transaction = getTransactionFromBytes(draft.transactionBytes);
-        groupItems.value.push({
+        groupItemsToAdd.push({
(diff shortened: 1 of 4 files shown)


front-end/src/renderer/stores/storeTransactionGroup.ts
```diff
@@ -100,6 +108,7 @@ const useTransactionGroupStore = defineStore('transactionGroup', () => {
    * @param index
    */
   function duplicateGroupItem(index: number) {
+    const lastItem = groupItems.value[groupItems.value.length - 1];
```
jbair06, line 111: Is there a deeper reason for adding the duplicated item to the end of the list as opposed to the adding it directly after the current item? There are potential some issues with either approach:

1. if adding the duplicate to the end, the newDate is based on the original item, and therefore the duplicate item might be added to the end but have a valid time less then previous items. 
2. if adding the duplicate just after the current item, the subsequent seq (and maybe even valid times) may need to be updated.

I think these issues might be addressed elsewhere, but should be considered. The usage of seq at all should also be considered as valid time may be all that is needed.

jbair06 (changes requested):

front-end/src/renderer/stores/storeTransactionGroup.ts
jbair06: After an offline conversation, we'll keep this change for the time being. This should be revisited when we're ready to discuss the need for the seq property.

jbair06 (approved):

Issue #1391, opened 2024-12-19
Remove step from importing a recovery phrase


SimiHunjan:
### Problem

When I click on import recovery phrase I get to another window that requires me to click a button to proceed. When I click recovery phrase it should just take me to the window for me to enter the recovery phrase right after. 

### Solution

<img width="1204" alt="Screenshot 2024-12-18 at 8 54 07 PM" src="https://github.com/user-attachments/assets/b1e3a062-ed36-4e36-8907-3acb4130f78c" />

Remove the above step. And go straight to clicking on this button:
<img width="1281" alt="Screenshot 2024-12-18 at 8 54 53 PM" src="https://github.com/user-attachments/assets/9fcd0c7a-d2a3-406f-8910-b230750d3364" />

PR #1394, opened 2024-12-19
fix: remove step from Import

icoxxx:
**Description**:
Removed the first step from Import (recovery phrase).

**Related issue(s)**:
#1391 

**Fixes**:
 I saw there is a check to skip steps if (user.recoveryPhrase). Moved it in onBeforeMount

### DIFF
--- a/front-end/src/renderer/pages/RestoreKey/RestoreKey.vue
+++ b/front-end/src/renderer/pages/RestoreKey/RestoreKey.vue
@@ -202,12 +202,6 @@ onBeforeUnmount(() => {
   user.recoveryPhrase = null;
 });
 
-onBeforeMount(() => {
-  if (user?.recoveryPhrase) {
-    step.value += 1;
-  }
-});
-
 /* Watchers */
 watch(index, () => (inputIndexInvalid.value = false));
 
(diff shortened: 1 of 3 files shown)


front-end/src/renderer/pages/RestoreKey/RestoreKey.vue
```diff
@@ -202,6 +202,12 @@ onBeforeUnmount(() => {
   user.recoveryPhrase = null;
 });
 
+onBeforeMount(() => {
```
jbair06, line 205: This is to skip the recovery phrase input step if the recovery phrase is still cached? Is that what we want? Or should we allow the user to see the cached phrase and use a different phrase if so desired?

jbair06 (changes requested):

front-end/src/renderer/pages/RestoreKey/RestoreKey.vue
jbair06: I see the logic was just moved from Step 1, good catch there. I'm just wondering if this is what we want the flow to be (auto skip the new Step 1 if phrase is cached)

SvetBorislavov: We won't skip the recovery step, we will only prefill the boxes

icoxxx: done

jbair06 (dismissed):

SvetBorislavov (approved):

PR #1404, opened 2024-12-20
feat: update transaction description field

### LINKED ISSUE #1403: Update the "Description" field for creating a transaction and viewing the transaction details
### Problem

The text box for entering the description is narrower compared to the other fields on the page. After the transaction is created, the displayed description text is placed inside a column that does not match the width of the input text box.

### Solution

Increase the size of the text table and apply that same size to the description box in the transaction details view.

### Alternatives

_No response_

SimiHunjan:
**Description**:

Update the description field when creating a transaction and viewing the transaction details.
**Related issue(s)**:

Fixes #1403 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/front-end/src/renderer/pages/TransactionDetails/TransactionDetails.vue
+++ b/front-end/src/renderer/pages/TransactionDetails/TransactionDetails.vue
@@ -237,7 +237,7 @@ const handleSign = async () => {
   }
 
   const personalPassword = getPassword(handleSign, {
-    subHeading: 'Enter your application password to decrypt your private key',
+    subHeading: 'Enter your application password to access your private key',
   });
   if (passwordModalOpened(personalPassword)) return;
 
@@ -312,7 +312,7 @@ const handleApprove = async (approved: boolean, showModal?: boolean) => {
     }
 
     const personalPassword = getPassword(callback, {
-      subHeading: 'Enter your application password to decrypt your private key',
+      subHeading: 'Enter your application password to access your private key',
     });
     if (passwordModalOpened(personalPassword)) return;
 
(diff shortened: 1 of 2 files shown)


front-end/src/renderer/pages/TransactionDetails/TransactionDetails.vue
```diff
@@ -528,6 +528,7 @@ const sectionHeadingClass = 'd-flex justify-content-between align-items-center';
 const detailItemLabelClass = 'text-micro text-semi-bold text-dark-blue';
 const detailItemValueClass = 'text-small mt-1';
 const commonColClass = 'col-6 col-lg-5 col-xl-4 col-xxl-3 overflow-hidden py-3';
+const descriptionColClass = 'col-12 col-lg-12 col-xl-12 col-xxl-12\'';
```
SvetBorislavov, line 531: There is no need for multiple col-12's
Bootstrap uses the min-width media queries so that you can leave only `col-12` and it will apply to any larger screens

SvetBorislavov: Then since it is only a single class, you don't need a variable for that

front-end/src/renderer/pages/TransactionDetails/TransactionDetails.vue
```diff
@@ -629,7 +630,9 @@ const cancel = 'Cancel';
                     "
                   >
                     <h4 :class="detailItemLabelClass">Description</h4>
-                    <p :class="detailItemValueClass" data-testid="p-description-field">
+                    <p :class="detailItemValueClass" data-testid="p-description-field"
+                       style="white-space: normal; word-break: break-word; overflow-wrap: break-word;"
```
SvetBorislavov, line 634: Inline styles are bad practice, consider adding the `text-break` to the `detailItemValueClass` and remove the inline styles

SvetBorislavov (changes requested):

SimiHunjan:
new PR due to signing issues: https://github.com/hashgraph/hedera-transaction-tool/pull/1409

Issue #1340, opened 2024-12-12
Reduce application start up time


SimiHunjan:
### Problem

Users have reported the application takes a long time to load. On the order of 30-60 seconds.

### Solution

Reduce the application load time.

Issue #1281, opened 2024-12-03
The login/register window should not be displayed when the application is still loading


SimiHunjan:
### Problem

When you open the application, sometimes it takes few seconds to fully load. When loading the login/register window is seen in the background. 

### Solution

The login/register should not be visable when the application is loading.

SimiHunjan:
If we have the splash screen we should also remove it from these views 
![IMG_4340](https://github.com/user-attachments/assets/3898101c-7f37-43cb-891e-08805a5a8b15)
![IMG_4335](https://github.com/user-attachments/assets/318ba4dd-f833-4960-8e3c-4d7fb9d5b12f)

SvetBorislavov:
TBD

PR #1417, opened 2025-01-02
feat: application loading optimizations

SvetBorislavov:
**Description**:
This pull request improves the application's loading process. Data for each connected organization is now fetched in the background, along with additional user data. As a result, the initial load time is reduced, allowing users to interact with the application while some components are still loading. Additionally, a background has been added to the loading modal backdrop in certain areas.

**Related issue(s)**:

Fixes #
#1281 
#1340


yiliev0 (approved):
LGTM

Issue #1208, opened 2024-11-19
Allow the application to create and sign a transaction and not submit it


SimiHunjan:
### Problem

There may be some use cases where a transaction is created and signed, but submitted by or used by another tool. The application should allow you to:

(1) Create a transaction
(2) Share the transaction for signing 
(3) Does not submit the transaction
(4) The transaction is exported from the tool (can be via API)

### Solution

Design a solution for this feature.

SimiHunjan:
Maybe at the transaction creation stage, the window has an option to select the transaction to not be submitted once signed and then the transaction does not get promoted to the execute stage. Once the transaction receives its signature(s) it then moves to history.

PR #1440, opened 2025-01-13
feat: allow creation of sign-only transactions and export them

SvetBorislavov:
**Description**:
This pull request:
- adds transaction statuses to create a transaction that won't be executed.
- adds an export button so that users can export transaction files and deserialize them in other applications

**Related issue(s)**:

Fixes #
#1208 

**Checklist**

- [X] Tested


yiliev0 (approved):

back-end/apps/api/src/transactions/transactions.controller.ts
jbair06, line 1: Is the idea to have the user create the transaction, then manually change it to sign only afterwards?

jbair06 (changes requested):

back-end/apps/api/src/transactions/transactions.controller.ts
jbair06: I think that this status may need to be unalterable once the transaction is created and sent to the backend.

jbair06: We will discuss this further and add it to the next issue being created around the fine-tuning of this feature.

Issue #896, opened 2024-09-10
Add valid start to Transaction Group Details


jbair06:
### Problem

If the creator is creating a transaction group with lots of transactions (or just takes a long enough time creating the group), some transactions could have expired before the group is made. The creator would need to go through all transactions to manually ensure the valid start is still valid, in this case. 

### Solution

* Add a valid start ticking (and editable) component to the Transaction Group Details page
* Creating the transaction group would then update each transaction that has a valid start time older than the group's valid start time
* Ensure that this functionality still separates each valid start time by a millisecond or so.

PR #1441, opened 2025-01-13
feat: valid group start

icoxxx:
**Description**:
Introduced logic to ensure that a valid group start time is set during the creation of group transactions, fixes for faulty if condition that was leading to unexpected behavior and a wrong callback function in TransactionGroupProcessor.

**Related issue(s)**:
#896


jbair06 (changes requested):
I tried to change the group's valid start. When I tried to edit a transaction, then go back to the group, the valid start had been reset.

SvetBorislavov (dismissed):

Issue #1478, opened 2025-01-22
Remember the size and position of windows after restart


steven-sheehy:
### Problem

When the app is started it is stretched to 90% width of my window. On a 49" monitor, that is one large window. I adjust the horizontal size manually to an appropriate level but on restart it defaults again to very wide.

### Solution

Remember the size and position of windows and restore them after restarting.

PR #1487, opened 2025-01-27
feat: add window persisting data

SvetBorislavov:
**Description**:
This pull request adds a window state persistent.
An important note is that if the user has no personal user, the state won't be saved

**Related issue(s)**:

Fixes #
#1478


yiliev0 (approved):
LGTM. 
**Note**: If a user with more than two monitors drags the application from the main screen to a secondary monitor with a larger resolution, maximizes it, and then restarts the application, it reopens on the secondary monitor but at the maximum resolution of the main screen.

Issue #1426, opened 2025-01-08
Add checksum and account ID support


jbair06:
Checksums should appear wherever an account is loaded or entered. 

For data entry, this requires that all text fields accept `0.0.1234`, `0.0.1234-abcde`, `1234`, or `1234-abcde`. In the latter two cases, the values should be converted to the proper form: `0.0.1234-abcde`. If the checksum was manually entered, it should be validated allowing for incorrect checksums to be rejected and an error shown to the user.

For data selection, the accounts in a drop down list should show the checksum along side the account IDs.

For all other locations where an account is displayed, the checksum would appear along side the account ID.

PR #1491, opened 2025-01-28
feat: add checksum and account id support

icoxxx:
**Description**:
Checksums now appear wherever an account is loaded or entered in the app.
For data entry (add account) - it adds the checksum on blur of the input, showed in format like: 0.0.1234-abcde

**Related issue(s)**:

#1426


yiliev0 (commented):
When building the app, I get ESLint warnings for unused variables: `watch` in `AccountIdInput.vue` and `useAccountId` in `AppAutoComplete.vue`. Could you check them out?

Issue #1467, opened 2025-01-18
Navigation issues with the Account chooser


jbair06:
When the account chooser drop down is visible and navigatable via the arrow keys, pressing the down or up arrows should not loop the list. Or, if desired, it can, but the drop down doesn't properly handle the auto scrolling for this. It does not always show the account that the account field displays as the current selected item.

Remove the looping of arrow key navigation on the Account drop down list.

jbair06:
related to #1427

Issue #1427, opened 2025-01-08
Drop down list selection improvements


jbair06:
Drop down lists need to have better methods for selecting the currently highlighted value. The enter or tab keys should select the currently highlighted value, dismiss the popup, and move to the next component. Currently, it appears to select the value if not selected, and try to submit the form.

The escape key should dismiss the drop down without replacing the user entered text. Then tab or enter should move to the next component.

jbair06:
In addition, for the auto completion feature of the drop down, pressing the right arrow should fill in the next digit of the suggested selection.

icoxxx:
Can we consider it done with the fix for #1467?

PR #1507, opened 2025-01-30
fix: list loop and auto scroll in AppAutoComplete

icoxxx:
**Description**:
Fix for dropdown in AppAutoComplete (now it loops and handles auto scrolling properly).

PS: Looping is not removed but rather fixed (bug in the code), can remove it if it is necessary.
This PR includes the changes requested in #1427 

**Related issue(s)**:
#1467

### DIFF
--- a/front-end/src/renderer/components/ui/AppAutoComplete.vue
+++ b/front-end/src/renderer/components/ui/AppAutoComplete.vue
@@ -1,5 +1,5 @@
 <script setup lang="ts">
-import { computed, nextTick, onBeforeUnmount, onMounted, ref } from 'vue';
+import { computed, nextTick, onBeforeUnmount, onMounted, ref, watchEffect } from 'vue';
 
 import AppInput from '@renderer/components/ui/AppInput.vue';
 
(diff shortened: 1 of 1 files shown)


front-end/src/renderer/components/ui/AppAutoComplete.vue
```diff
@@ -135,8 +149,10 @@ function setValue(value: string) {
 }
 
 function scrollToItem(index: number) {
+  const correctId = filteredItems.value[index];
```
jbair06, line 152: I don't fully understand what this change does. Perhaps the name 'filteredItems' is left over from when the list was actually filtered based on user input? Can you explain why this 'correctIndex' change was needed?

jbair06 (changes requested):

SvetBorislavov (commented):
LGTM

SvetBorislavov:
@jbair06  what do you think

front-end/src/renderer/components/ui/AppAutoComplete.vue
icoxxx: In that func itemRefs.value[index] !== filteredItems.value[index] which was messing with the scroll

icoxxx: Generally, itemRefs was not populated correctly, but it can be done differently to make it more understandable. I fixed it in the last commit.

jbair06 (approved):

PR #1511, opened 2025-01-31
feat: add resend email for signers

### LINKED ISSUE #1336: Add a remind button to send out transaction notification emails to those users who have not signed the transaction
### Problem

Transactions can be scheduled for some time into the future. After a few days you may want to remind the users who have not yet signed the transaction. If we do that today, we have to compile a list of those users who did not sign, create an email, and send it off. It would be easier if there was a way in the app to click a button to remind the users who have not yet signed the transaction. For a future case, text message communication directly to users phones would be a neat option too.

### Solution

See above

### Alternatives

_No response_

SvetBorislavov:
**Description**:
This pull request adds a manual reminder for the required singers of a transaction.
There is also a rate limit that limits the reminder to one per minute

**Related issue(s)**:

Fixes #1336


jbair06 (changes requested):

jbair06 (approved):
LGTM

Issue #1458, opened 2025-01-15
Transactions in a group need a visual indication of sign status


jbair06:
Transaction groups can have one or many transactions. If one or more of the transactions are signed, and one or more are not, the group will appear in the 'Ready to Review' and 'In Progress' tabs without any visual indication as to which transaction is signed and which is not. The user can go into each transaction in the group and see if it is signed, but a visual indication in the Transaction Group Page would be more helpful.

Add some visual indication in the Transaction Group Page whether a transaction is signed, or not. This is only needed in the 'Ready to Sign' or 'In Progress' pages.

PR #1512, opened 2025-01-31
feat: indicator for unsigned transactions within a group

icoxxx:
**Description**:
Implemented logic to handle notifications for unsigned transactions within a group. Added a visual indicator displaying the number of transactions awaiting signatures in the ReadyToSign and InProgress components, along with an indicator  (exclamation mark) in the TransactionGroupDetails page.

**Related issue(s)**:
#1458


jbair06 (changes requested):
I like this, overall. I also like the counter in the 'Ready to Sign' and 'In Progress' pages indicating how many need to be signed. Good idea. 

Having played with it for a bit, I think it should be slightly adjusted. Maybe we can make it a bit more subtle. How would it look if the buttons were given a red border if unsigned instead of the '!'? 

And for the number indicator for how many need to be signed, move it from the far left and put it in the button like where the '!' is.

icoxxx:
> I like this, overall. I also like the counter in the 'Ready to Sign' and 'In Progress' pages indicating how many need to be signed. Good idea.
> 
> Having played with it for a bit, I think it should be slightly adjusted. Maybe we can make it a bit more subtle. How would it look if the buttons were given a red border if unsigned instead of the '!'?
> 
> And for the number indicator for how many need to be signed, move it from the far left and put it in the button like where the '!' is.

Yep, that looks better in terms of both design and code. Do you want the buttons displaying the count of unsigned transactions to have a red border as well, or only on the page showing transactions from the group (as it is now)?

jbair06:
> Yep, that looks better in terms of both design and code. Do you want the buttons displaying the count of unsigned transactions to have a red border as well, or only on the page showing transactions from the group (as it is now)?

Can you provide a screen shot of both? I think as it is now would be good, but it would be helpful to see.

icoxxx:
> > Yep, that looks better in terms of both design and code. Do you want the buttons displaying the count of unsigned transactions to have a red border as well, or only on the page showing transactions from the group (as it is now)?
> 
> Can you provide a screen shot of both? I think as it is now would be good, but it would be helpful to see.

jbair06:
I showed this to others, to get more eyes. The feedback I got was that the red is more concerning than informative. 

After further discussion, the direction we want to try is undo these changes and instead separate the signed and unsigned transactions in some manner. This could be done using tabs, or some filtering method. 

Another reason for this approach is to make the information we're trying to convey to be more intuitive, and not rely on documentation to explain the feature. 

This also includes the number badge indicator, it won't be required, even as helpful as it is.

icoxxx:
> I showed this to others, to get more eyes. The feedback I got was that the red is more concerning than informative.
> 
> After further discussion, the direction we want to try is undo these changes and instead separate the signed and unsigned transactions in some manner. This could be done using tabs, or some filtering method.
> 
> Another reason for this approach is to make the information we're trying to convey to be more intuitive, and not rely on documentation to explain the feature.
> 
> This also includes the number badge indicator, it won't be required, even as helpful as it is.

Okay, I kept the design simple and added a filter for unsigned transactions, or a notification instead if all transactions are unsigned. At first, I thought about using a dropdown, but then I wondered—if i were the user would I really need one when there are only two options (all / unsigned)....

jbair06:
I like this idea. Much cleaner! Let's see if we can make it even simpler: If the details page is viewed from the 'Ready to Sign' or 'In Progress' it should show only the unsigned transactions by default. the check box should then say 'Show All' and by default be unchecked. If the details page is viewed from 'History' or 'Ready to Execute', then it should just show all transactions and not have the check box.

This means you can also remove the 'All transactions are unsigned' bit. What do you think?

icoxxx:
> I like this idea. Much cleaner! Let's see if we can make it even simpler: If the details page is viewed from the 'Ready to Sign' or 'In Progress' it should show only the unsigned transactions by default. the check box should then say 'Show All' and by default be unchecked. If the details page is viewed from 'History' or 'Ready to Execute', then it should just show all transactions and not have the check box.
> 
> This means you can also remove the 'All transactions are unsigned' bit. What do you think?

@jbair06 
As far as I know, TransactionGroupDetails can't be viewed from 'History' or 'Ready to Execute' but they rather open the single transaction details...? Am I wrong? If that's the case, we don't need to track the previous route.
I've included the changes, and now it shows unsigned transactions by default.

jbair06:
You are correct about 'History' and 'Ready to execute'. We may want to revisit how the transaction groups show up in 'Ready to Execute' though. If a group is 'atomic' or 'sequential', then they should only be 'ready to execute' once all are signed, and so should probably show up as a group. I'll add this to the 'atomic' transaction group issue.

front-end/src/renderer/utils/transactionSignatureModels/index.ts
```diff
@@ -45,66 +45,115 @@ export const getSignatureEntities = (transaction: Transaction) => {
...
+
+  const signedSigners = new Set(Object.keys(transaction.getSignatures()));
+
+  const isFullySigned = [...allRequiredSigners].every(pk => signedSigners.has(pk));
+
+  return isFullySigned;
+};
+
 /* Returns whether a user should sign the transaction */
 export const publicRequiredToSign = async (
```
jbair06, line 98: I think these changes seemed to have broken a few other things. Specifically, the TransactionDetailsHeader, after signing a transaction, the 'sign' button stays.

jbair06 (changes requested):

front-end/src/renderer/pages/TransactionDetails/TransactionDetails.vue
```diff
@@ -102,9 +106,27 @@ async function fetchTransaction(id: string | number) {
...
       );
       transactionBytes = hexToUint8Array(orgTransaction.value.transactionBytes);
+
+      if (orgTransaction.value?.groupItem.groupId) {
+        if (user.selectedOrganization?.serverUrl) {
+          const orgGroup = await getApiGroupById(
+            user.selectedOrganization?.serverUrl,
+            orgTransaction.value.groupItem.groupId,
+          );
+          orgGroupDescription.value = orgGroup.description;
```
SvetBorislavov, line 116: You can use only one state variable. for example only groupDescription as It is a basic string, no need to have two states

SvetBorislavov (changes requested):

front-end/src/renderer/utils/transactionSignatureModels/index.ts
icoxxx: This is a leftover and will not use it. Deleted it in the last commit. The problem occurred when rebased with main, will investigate it

icoxxx: Resolved

jbair06:
I played around with this a bit. I noticed an issue. When, as the creator, I go to the 'in progress' section, and open a transaction group, I see what appears to be all transactions are checked off as being signed ONLY if they have received ALL signatures. This includes transactions that have thresholds set up in such a way that not all signatures are actually required. Then, I clicked into a transaction, any transaction, and clicked back and was then able to see all transactions that have valid signatures now checked off. 

Screen Recording 2025-02-13 at 3 03 17 PM

icoxxx:
> I played around with this a bit. I noticed an issue. When, as the creator, I go to the 'in progress' section, and open a transaction group, I see what appears to be all transactions are checked off as being signed ONLY if they have received ALL signatures. This includes transactions that have thresholds set up in such a way that not all signatures are actually required. Then, I clicked into a transaction, any transaction, and clicked back and was then able to see all transactions that have valid signatures now checked off.
> 
> Screen Recording 2025-02-13 at 3 03 17 PM [ Screen Recording 2025-02-13 at 3 03 17 PM ](https://private-user-images.githubusercontent.com/117694970/413094281-742d4479-19cb-432a-b808-4cb1eba61df6.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3Mzk1NDY4NjIsIm5iZiI6MTczOTU0NjU2MiwicGF0aCI6Ii8xMTc2OTQ5NzAvNDEzMDk0MjgxLTc0MmQ0NDc5LTE5Y2ItNDMyYS1iODA4LTRjYjFlYmE2MWRmNi5naWY_WC1BbXotQWxnb3JpdGhtPUFXUzQtSE1BQy1TSEEyNTYmWC1BbXotQ3JlZGVudGlhbD1BS0lBVkNPRFlMU0E1M1BRSzRaQSUyRjIwMjUwMjE0JTJGdXMtZWFzdC0xJTJGczMlMkZhd3M0X3JlcXVlc3QmWC1BbXotRGF0ZT0yMDI1MDIxNFQxNTIyNDJaJlgtQW16LUV4cGlyZXM9MzAwJlgtQW16LVNpZ25hdHVyZT04NThiNzE4YjcxZGI1N2Y2YzM3NWU3ZDI5ODhmY2M0YjZjZjlmNmM5MGUwNjIwNWM0ZTVkOTU1OTZmY2IxYjk1JlgtQW16LVNpZ25lZEhlYWRlcnM9aG9zdCJ9.eYGeiYwnffKM51INPZitly81MFXI0hjbT8ui6biOLRc) [ ](https://private-user-images.githubusercontent.com/117694970/413094281-742d4479-19cb-432a-b808-4cb1eba61df6.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3Mzk1NDY4NjIsIm5iZiI6MTczOTU0NjU2MiwicGF0aCI6Ii8xMTc2OTQ5NzAvNDEzMDk0MjgxLTc0MmQ0NDc5LTE5Y2ItNDMyYS1iODA4LTRjYjFlYmE2MWRmNi5naWY_WC1BbXotQWxnb3JpdGhtPUFXUzQtSE1BQy1TSEEyNTYmWC1BbXotQ3JlZGVudGlhbD1BS0lBVkNPRFlMU0E1M1BRSzRaQSUyRjIwMjUwMjE0JTJGdXMtZWFzdC0xJTJGczMlMkZhd3M0X3JlcXVlc3QmWC1BbXotRGF0ZT0yMDI1MDIxNFQxNTIyNDJaJlgtQW16LUV4cGlyZXM9MzAwJlgtQW16LVNpZ25hdHVyZT04NThiNzE4YjcxZGI1N2Y2YzM3NWU3ZDI5ODhmY2M0YjZjZjlmNmM5MGUwNjIwNWM0ZTVkOTU1OTZmY2IxYjk1JlgtQW16LVNpZ25lZEhlYWRlcnM9aG9zdCJ9.eYGeiYwnffKM51INPZitly81MFXI0hjbT8ui6biOLRc)

Fixed and synced with main

jbair06 (approved):
Manual tests all look great!

SvetBorislavov (approved):

Issue #1459, opened 2025-01-15
Nicknames should appear before object it names


jbair06:
Whether an account, or key, or whatever entity, if it has a nickname then the nickname should appear first, with the entity in parentheses. If the user is nicknaming an entity, they likely do this due to the entity name having no real meaning to them.

Put nickname first, with entity name in parentheses.

PR #1514, opened 2025-02-03
feat: nickname before object it names

icoxxx:
**Description**:
This update ensures that whenever an entity has a nickname, the nickname is displayed first, followed by the entity name in parentheses.
+++Edit: Improved performance by calling 'getAll' (from '@renderer/services/accountsService') to retrieve user's linked accounts only in TransactionDetails and pass the result as props to child components.

**Related issue(s)**:
#1459
 
Fixes #
Overflow in ContactList and Accounts pages

### DIFF
--- a/front-end/src/renderer/pages/TransactionDetails/TransactionDetails.vue
+++ b/front-end/src/renderer/pages/TransactionDetails/TransactionDetails.vue
@@ -1,6 +1,6 @@
 <script setup lang="ts">
 import type { Transaction } from '@prisma/client';
-import type { ITransactionFull, IUserLinkedAccounts } from '@main/shared/interfaces';
+import type { ITransactionFull } from '@main/shared/interfaces';
 
 import { computed, onBeforeMount, ref, watch } from 'vue';
 import { useRouter } from 'vue-router';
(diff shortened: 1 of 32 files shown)


front-end/src/renderer/utils/index.ts
```diff
@@ -106,3 +112,36 @@ export const throwError = (errorMessage: string) => {
 
 export const getErrorMessage = (error: unknown, defaultErrorMessage: string) =>
   error instanceof Error ? error.message : defaultErrorMessage;
+
+export function handleFormatAccount(
+  allAccounts: HederaAccount[] | undefined,
+  accountToCheck: AccountInfo | HederaAccount,
+): string {
+  const accountData = useAccountId();
```
SvetBorislavov, line 120: This way you will make a request to the mirror node everytime, consider extracting the logic for formatting the checksum out of the useAccountId and use it directly

front-end/src/renderer/pages/TransactionDetails/TransactionDetails.vue
```diff
@@ -166,6 +182,15 @@ onBeforeMount(async () => {
     ),
   ]);
   nextId.value = result[1];
+
+  if (isUserLoggedIn(user.personal)) {
+    allLinkedAccounts.value = await getAll({
```
SvetBorislavov, line 187: Why are you fetching all the accounts instead of only the required account ids in the proper pages?

SvetBorislavov (changes requested):

SvetBorislavov (approved):

Issue #1375, opened 2024-12-17
Clear button does not clear manually typed words in Recovery Phrase input fields


yiliev0:
### Description

![image](https://github.com/user-attachments/assets/d3e2f55b-f842-4b83-b38d-262de826b950)

### Steps to reproduce

1. Navigate to the Account Setup page.
2. In the Recovery Phrase section, manually type words into the input fields (e.g., "test", "test", etc.).
3. Click the Clear button.
4. Observe the input fields.

### Additional context

Expected result: All input fields should be cleared when clicking the Clear button, regardless of how the words were entered.

Actual result: Words manually typed into the input fields are not cleared after clicking the Clear button.

### Version

0.5.0

### Operating system

None

PR #1515, opened 2025-02-04
fix: clear words in RestoreKey

icoxxx:
**Description**:
Simple fix of clear words handler in RestoreKey. (watcher in child component wasn't detecting the changes on clicking of the 'clear' button).

**Related issue(s)**:
#1375

### DIFF
--- a/front-end/src/renderer/components/RecoveryPhrase/Import.vue
+++ b/front-end/src/renderer/components/RecoveryPhrase/Import.vue
@@ -100,15 +100,6 @@ watch(words, async newWords => {
   }
 });
 
-watch(
-  () => user.recoveryPhrase,
-  async newRecoveryPhrase => {
-    if (!newRecoveryPhrase) {
-      handleClearWords();
-    }
-  },
-);
-
 watchEffect(() => {
   if (props.shouldClear) {
     handleClearWords();
--- a/front-end/src/renderer/components/RecoveryPhrase/Import.vue
+++ b/front-end/src/renderer/components/RecoveryPhrase/Import.vue
@@ -1,11 +1,19 @@
 <script setup lang="ts">
-import { onBeforeMount, ref, watch } from 'vue';
+import { onBeforeMount, ref, watch, watchEffect } from 'vue';
 
 import useUserStore from '@renderer/stores/storeUser';
 
 import { validateMnemonic } from '@renderer/services/keyPairService';
 
 import AppRecoveryPhraseWord from '@renderer/components/ui/AppRecoveryPhraseWord.vue';
+
+/* Props */
+const props = defineProps<{
+  shouldClear?: boolean;
+}>();
+
+/* Emits */
+const emit = defineEmits(['reset-cleared']);
 
 /* Constants */
 const WORD_COUNT = 24;
@@ -100,6 +108,13 @@ watch(
     }
   },
 );
+
+watchEffect(() => {
+  if (props.shouldClear && inputRefs.value && inputRefs.value.length > 0) {
+    handleClearWords();
+    emit('reset-cleared', false);
+  }
+});
 </script>
 <template>
   <div>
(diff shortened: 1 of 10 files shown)


SvetBorislavov (changes requested):
The fix is not applied everywhere.
To apply it, change the setter function the userStore to this:
```
  const setRecoveryPhrase = async (words: string[] | null) => {
    if (words === null) {
      recoveryPhrase.value = { words: [], hash: '' } as unknown as RecoveryPhrase;
      await nextTick();
      recoveryPhrase.value = null;
    } else {
      recoveryPhrase.value = await ush.createRecoveryPhrase(words);
    }
  };
  ```

and find everywhere in the app the direct assignments of the recovery phrase and replace them with the setter function

File: front-end/src/renderer/stores/storeUser.ts [hunk 1]
```diff
@@ -109,6 +109,8 @@ const useUserStore = defineStore('user', () => {
   /** Keys */
   const setRecoveryPhrase = async (words: string[] | null) => {
     if (words === null) {
+      recoveryPhrase.value = { words: [], hash: '' } as unknown as RecoveryPhrase;
```
jbair06, line 112: While this does work, I'm not a huge fan of it. The underlying issue appears to be that storeUser.recoveryPhrase is only set when the phrase entered (either typed, or pasted, or generated) is a valid mnemonic. So when the user enters in only one, or a few, or changes some of the words and the result is no longer a valid mnemonic, the recoveryPhrase is never set. This means that when 'clear' is pressed, the recoveryPhrase is set to null, but was already null, so no event is fired, nor caught. So the Import component never processes this event. 

As you likely understood this, you just forced the change to trigger the event. Would it be a bit better to use provide/inject to have parents of the Import component provide the 'clear' event, and the Import component can just clear the words on its own? 

If we were to make this change, it is likely still good to keep the watch on storeUser.recoveryPhrase for any other case outside of the 'clear' button. Though, I wonder if a deeper look into when the recoveryPhrase is actually set would be good, too. For example, it is set on submit and a few other places, does it need to be set as soon as the words are valid?

jbair06 (changes requested):

front-end/src/renderer/stores/storeUser.ts
icoxxx: > While this does work, I'm not a huge fan of it. The underlying issue appears to be that storeUser.recoveryPhrase is only set when the phrase entered (either typed, or pasted, or generated) is a valid mnemonic. So when the user enters in only one, or a few, or changes some of the words and the result is no longer a valid mnemonic, the recoveryPhrase is never set. This means that when 'clear' is pressed, the recoveryPhrase is set to null, but was already null, so no event is fired, nor caught. So the Import component never processes this event.
> 
> As you likely understood this, you just forced the change to trigger the event. Would it be a bit better to use provide/inject to have parents of the Import component provide the 'clear' event, and the Import component can just clear the words on its own?
> 
> If we were to make this change, it is likely still good to keep the watch on storeUser.recoveryPhrase for any other case outside of the 'clear' button. Though, I wonder if a deeper look into when the recoveryPhrase is actually set would be good, too. For example, it is set on submit and a few other places, does it need to be set as soon as the words are valid?

Yes, it was indeed intended to trigger the event. I redid it with something similar to what you suggested, although the previous approach wasmuch cleaner.

yiliev0 (dismissed):
LGTM, tested.

front-end/src/renderer/components/RecoveryPhrase/Import.vue
```diff
@@ -100,6 +108,13 @@ watch(
     }
   },
 );
+
+watchEffect(() => {
```
jbair06, line 112: shouldClear needs to be watched, but why inputRefs? This causes manual altering of any one of the words to clear all the words, I think.

jbair06 (changes requested):

jbair06:
@yiliev0 We may need to adjust the tests to catch for the issue I described in my review.

SvetBorislavov (approved):

Issue #1464, opened 2025-01-17
Selecting new date for a transaction should clear the time


jbair06:
When the user is selecting a future date for a transaction, the time portion doesn't reset, or change. This means that whatever the time showed before, will still be the same. Generally, the user will want to specify the time for the transaction. The experience noted by some users is that they don't bother with the seconds portion. This leaves a random second value. 

The tool should reset the time portion when a new date is selected for a more clean user experience. This is solely for visual cleanliness.

PR #1517, opened 2025-02-04
fix: clear time in RuningClockDatePicker

icoxxx:
**Description**:
Reset time to default value (00:00) on selecting a new date in RuningClockDatePicker component

**Related issue(s)**:
#1464

### DIFF
--- a/front-end/src/renderer/components/RunningClockDatePicker.vue
+++ b/front-end/src/renderer/components/RunningClockDatePicker.vue
@@ -1,7 +1,7 @@
 <script setup lang="ts">
 import type { VueDatePickerProps } from '@vuepic/vue-datepicker';
 
-import { onMounted, onUnmounted, ref } from 'vue';
+import { onMounted, onUnmounted, ref, watch } from 'vue';
 
 import AppDatePicker from '@renderer/components/ui/AppDatePicker.vue';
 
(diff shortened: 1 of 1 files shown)


SvetBorislavov (approved):

front-end/src/renderer/components/RunningClockDatePicker.vue
```diff
@@ -35,6 +35,22 @@ function stopInterval() {
...
 }
 
+function resetTime(date: Date) {
+  const newDate = new Date(date);
+  newDate.setHours(0, 0, 0, 0);
+  return newDate;
+}
+
+function handleUpdateValidStart(date: Date) {
+  const isNewDate = date.toDateString() !== props.modelValue.toDateString();
```
jbair06, line 45: If the date selected is today's date (whether using the 'now' button, or manually selecting today), then 'isNewDate' should be false. While the RunningClockDatePicker will see the time is past and update it to the current time, visually it is a bit jarring for the user to see they selected today, and the time is 00:00:00.

jbair06 (changes requested):

jbair06:
I've created issue #1529 that has a couple of improvements that should be added to this PR, if possible. This UX improvements will help make it smoother.

jbair06:
Commit  https://github.com/hashgraph/hedera-transaction-tool/pull/1517/commits/b17b2ec611e28e8c423d474b3414bb28244a904d fixes #1529

jbair06 (approved):

yiliev0 (approved):
LGTM, tested.

PR #1528, opened 2025-02-06
fix: user keys fetching

### LINKED ISSUE #1149: user keys didn't repull on every select
### Description

While looking at the contact list, a user added a new key to the organization. I was not able to refresh the view, even by selecting another user and reselecting the user with the new key.

### Steps to reproduce

1. Select an organization
2. Select Contact List
3. Select a Contact
4. Have that contact upload a new key
5. Verify that the key is not visible in your app
6. Select another contact, then back to the first
7. Verify that the key is still not visible in your app
8. Select another tab (like Transaction) then back to Contact List and select the contact again
9. Verify that the key is now visible

### Additional context

_No response_

### Hedera network

_No response_

### Version

v0.4.1

### Operating system

macOS

SvetBorislavov:
**Description**:
This pull request updates the fetching of the user keys. It is now paginated and prefetched whenever a user is selected in the contact list

**Related issue(s)**:

Fixes #1149

### DIFF
--- a/back-end/apps/api/src/user-keys/user-keys-all.controller.spec.ts
+++ b/back-end/apps/api/src/user-keys/user-keys-all.controller.spec.ts
@@ -1,0 +1,54 @@
+import { Test, TestingModule } from '@nestjs/testing';
+import { mockDeep } from 'jest-mock-extended';
+
+import { BlacklistService, guardMock } from '@app/common';
+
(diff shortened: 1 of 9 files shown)


jbair06 (approved):
LGTM, lazy loading is working great in my tests!

PR #1532, opened 2025-02-11
feat: allow skipping account setup

### LINKED ISSUE #1307: Revisit public key requirement for backend
Currently a user that connects to the backend is required to have a public key uploaded. Without one, it takes them back to the setup process. In the future, when observers are fully implemented and used, we may have users that will not have need of public keys on the backend. 

We need to revisit this requirement and determine:
1. Is it still needed
2. What effort would be required to remove it
3. If it is needed, can we circumvent that need in the case of observers

SvetBorislavov:
**Description**:
This pull request adds a skip button in the account setup page so that organizational users can skip the account setup

**Related issue(s)**:

Fixes #1307


yiliev0 (approved):
LGTM

Issue #1542, opened 2025-02-12
Complex Key nicknames don't always save correctly


Sheng-Long:
### Description

Complex Key nicknames don't always save correctly

### Steps to reproduce

Try to save complex key nicknames in the edit mode

### Version

current

### Operating system

None

jbair06:
Can no longer replicate issue.

PR #1544, opened 2025-02-12
1342 threshold nicknames

### LINKED ISSUE #1342: Create nicknames for thresholds
### Problem

Thresholds can often represent multiple groups/teams within an organization. To quickly understand which team or group has completed their signing requirements, it would be easier for users to reference the threshold group by a given nickname. 

### Solution

Add threshold group nicknames

### Alternatives

_No response_

Sheng-Long:
**Description**:
fixes threshold nickname bugs

**Related issue(s)**:

Fixes #1342, #1542 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/front-end/src/main/services/localUser/complexKeys.ts
+++ b/front-end/src/main/services/localUser/complexKeys.ts
@@ -90,10 +90,13 @@ export const deleteComplexKey = async (id: string) => {
   });
 };
 
-export const updateComplexKey = async (id: string, newKeyListBytes: Uint8Array, nickname?: string) => {
+export const updateComplexKey = async (id: string, newKeyListBytes?: Uint8Array, nickname?: string) => {
   const prisma = getPrismaClient();
+  let newProtobufEncoded;
 
-  const newProtobufEncoded = Buffer.from(newKeyListBytes).toString('hex');
+  if (newKeyListBytes) {
+    newProtobufEncoded = Buffer.from(newKeyListBytes).toString('hex');
+  }
 
   await prisma.complexKey.updateMany({
     where: {
--- a/front-end/src/preload/localUser/complexKeys.ts
+++ b/front-end/src/preload/localUser/complexKeys.ts
@@ -11,7 +11,7 @@ export default {
     getComplexKey: (userId: string, keyListBytes: Uint8Array): Promise<ComplexKey> =>
       ipcRenderer.invoke('complexKeys:getComplexKey', userId, keyListBytes),
     delete: (id: string): Promise<ComplexKey[]> => ipcRenderer.invoke('complexKeys:delete', id),
-    update: (id: string, newKeyListBytes: Uint8Array, nickname?: string): Promise<ComplexKey> =>
+    update: (id: string, newKeyListBytes?: Uint8Array, nickname?: string): Promise<ComplexKey> =>
       ipcRenderer.invoke('complexKeys:update', id, newKeyListBytes, nickname),
   },
 };
--- a/front-end/src/renderer/components/ComplexKey/ComplexKey.vue
+++ b/front-end/src/renderer/components/ComplexKey/ComplexKey.vue
@@ -1,5 +1,5 @@
 <script setup lang="ts">
-import { onMounted, ref } from 'vue';
+import { onBeforeMount, ref } from 'vue';
 import { Key, KeyList } from '@hashgraph/sdk';
 
 import ComplexKeyThreshold from '@renderer/components/ComplexKey/ComplexKeyThreshold.vue';
@@ -10,7 +10,7 @@ const props = defineProps<{
 }>();
 
 /* Emits */
-const emit = defineEmits(['update:modelKey', 'update:nickname']);
+const emit = defineEmits(['update:modelKey']);
 
 /* State */
 const keyList = ref<KeyList>(new KeyList());
@@ -25,12 +25,8 @@ const handleKeyListRemove = () => {
   emit('update:modelKey', keyList.value);
 };
 
-function handleUpdateNickname(nickname: string, keyId: string, keyList: KeyList) {
-  emit('update:nickname', nickname, keyId, keyList);
-}
-
 /* Hooks */
-onMounted(() => {
+onBeforeMount(() => {
   if (props.modelKey instanceof KeyList) {
     keyList.value = props.modelKey;
   }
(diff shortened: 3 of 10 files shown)


yiliev0 (commented):
```
[plugin:vite-plugin-eslint] [plugin vite-plugin-eslint] src/renderer/components/ComplexKey/ComplexKeyModal.vue: 
/Users/yordaniliev/Desktop/hedera/3/hedera-transaction-tool/front-end/src/renderer/components/ComplexKey/ComplexKeyModal.vue
   6:10  warning  'encodeKey' is defined but never used       @typescript-eslint/no-unused-vars
   6:37  warning  'isUserLoggedIn' is defined but never used  @typescript-eslint/no-unused-vars
  28:7   warning  'user' is assigned a value but never used   @typescript-eslint/no-unused-vars
```

yiliev0 (changes requested):
Selecting the Complex tab in Create Account transactions throws a "useUserStore is not defined" error toast message and the "+Create new" button fails to open the modal. 
image

yiliev0:
Issue#1 
When saving a complex key structure and returning to it, the threshold names are not persisted.
Steps to reproduce:
1. Navigate to create account tx
2. Click on +Create new button
3. Create a threshold with two inner thresholds inside
4. Set nicknames to thresholds
5. click on "Save Complex Key" name and set name to this structure
6. Navigate back to the transaction or create a new one, instead, click on Add existing complex structure and select the structure from step 5.
7. Observe the missing nicknames


Issue#2
When creating a transaction with thresholds that have nicknames, the nicknames are not saved and displayed in the transaction details
1. Navigate to create account tx
2. Click on +Create new button
3. Create a threshold with two inner thresholds inside
4. Set nicknames to thresholds
5. Sign & Execute the transaction
6. Once navigated to the Transaction Details, under the account create info section, click on "See Details"
7. Observe the default "Threshold" nicknames

Sheng-Long:
New changes should have fixed the issues above

yiliev0:
@Sheng-Long Issues are still visible. 

1. Complex key now can not be saved, error is thrown:

```
Error occurred in handler for 'complexKeys:add': Error: Complex key already exists!
    at addComplexKey (/Users/yordaniliev/Desktop/hedera/3/hedera-transaction-tool/front-end/dist-electron/main/index.js:1810:11)
    at async WebContents.<anonymous> (node:electron/js2c/browser_init:2:87023)
```


2. After transaction execution, structure nicknames are not displayed properly, check screenshots:
image
image

jbair06:
Here are a few additions/changes that we've discussed:

1. The top level threshold needs to look different from the rest as it is the 'base' threshold.
  a. it should not be collapsible
  b. needs to more resemble the concept of a base level
  c. the 'x' to remove should be a 'clear' or 'reset' button as it doesn't remove the base threshold but instead clears the complex key
  d. the nickname of the base threshold should have placeholder text 'Complex Key Nickname'.
2. All sub thresholds placeholder text should be 'Threshold Nickname'
3. All thresholds will be saved as complex keys, but only the 'base' threshold should appear in the saved key list. In the local database, complex keys will need to be marked such that only base keys appear. Possible solutions:
  a. add 'sub_key', 'hidden', or other such flag to indicate that which 'complex' keys appear
  b. add 'parent_id' column to tie each complex key with its parent. 
4. Only use a single button to save/finish. Whether a dynamic 'save'/'done' or just a single 'save' button remove unneeded buttons

jbair06:
Some addition concerns:
1. should nicknames be unique?
2. should key structures (protobufEncoded) be unique?
3. is an 'index' field helpful when saving the sub keys?
4. should the tool auto fill the nickname when the user creates a key structure that matches one already saved?

jbair06:
> should nicknames be unique?
I see no benefit to having nicknames be strictly unique. Having two nicknames be the same in the same complex key structure would not be ideal, but should let the user decide on what they want to do.

> should key structures (protobufEncoded) be unique?
The main purpose behind making the protobufEncoded be unique is to allow the app to try and determine the nickname of a structure when displaying the complex key throughout the app. The app pulls the current key (or key list) from the network when displaying account info. It then tries to match that key structure to one saved in the complex keys. If it finds a match, it can display the nickname. In order for this to work as described, the key structures need to be unique. 

However, it may be possible to match the full complex key instead of individual sub keys. If the full complex key is matched (and parent_id is used to link sub keys to the complex key), then the nicknames can be displayed for the entire structure. If no complex key matches, then no nicknames will be displayed. This could result in nicknames being 'lost' when the account key is altered as the full complex key will no longer match.

If, in addition to parent_id, we also link the complex key to an account (on account create/update, or when altering the nicknames of an account's complex key), then we would be able to still retain the nicknames even if the key structure were to change. In the event of a change, the nickname of the sub key that changed would need to either be removed, or a lot of extra logic would be needed to try and best guess the name of the changed sub keys based on the previous key structure. This could be a many-to-one relationship, as multiple accounts can use the same key structure. In this case, the structure no longer needs to be unique. 

> is an 'index' field helpful when saving the sub keys?
Probably not. The concept of an index is actually somewhat problematic. The key structure returned from the network is not actually guaranteed to be in the same order. This is due to potential changes made the the order during serializing/deserializing in the SDK or even translation from the network to mirror node. While unlikely, it is possible. This causes the above point to be even more difficult to solve, as it becomes a bit more work to match if we need to account for possible changes in order.

> should the tool auto fill the nickname when the user creates a key structure that matches one already saved?
Though referenced in #2, this is called out to emphasize the need to ensure user experience is not unexpected as that is the main goal. If a user is creating a key, and the structure gets auto matched to an existing key saved locally, auto filling the nickname might confuse users.

jbair06:
> The top level threshold needs to look different from the rest as it is the 'base' threshold.
> a. it should not be collapsible
> b. needs to more resemble the concept of a base level
> c. the 'x' to remove should be a 'clear' or 'reset' button as it doesn't remove the base threshold but instead clears the complex key
> d. the nickname of the base threshold should have placeholder text 'Complex Key Nickname'.
> All sub thresholds placeholder text should be 'Threshold Nickname'
> All thresholds will be saved as complex keys, but only the 'base' threshold should appear in the saved key list. In the local database, complex keys will need to be marked such that only base keys appear. Possible solutions:
> a. add 'sub_key', 'hidden', or other such flag to indicate that which 'complex' keys appear
> b. add 'parent_id' column to tie each complex key with its parent.
> Only use a single button to save/finish. Whether a dynamic 'save'/'done' or just a single 'save' button remove unneeded buttons

Do this, go with 3a (go with sub_key). 
Nicknames do not need to be unique. 
Key structures also do not need to be unique. When viewing an account key, the entire complex key can be matched, but the sub keys in a complex key will not need to be matched. This allows the same key structure to be used in many complex keys, but have different nicknames because they are in essence different keys.

jbair06:
After further discussion, this feature will need to be redesigned if it is required, due to the amount of edge cases it will need to handle.

Issue #1547, opened 2025-02-13
Draft modal UI regression: missing explanation text and icon


yiliev0:
### Description

Previously, when leaving the transaction page, users were presented with a draft modal that included explanatory text and an icon to clearly prompt them to save or discard their draft changes. The current implementation has regressed to a minimalistic "Save Edits" window without the additional context. 

![Image](https://github.com/user-attachments/assets/dfad561a-c4ba-47c5-b58b-301c50d3ffcf)

![Image](https://github.com/user-attachments/assets/aa55660a-7f6c-44b6-ac0e-16b1f0b5484c)

### Steps to reproduce

Attempt to leave a transaction page with some changes.

### Version

0.9.0

### Operating system

None

PR #1548, opened 2025-02-13
fix: explanatory text & icon in modal, fix: skip modal on logout

icoxxx:
**Description**:
This PR adds the missing explanatory text and icon in modals, along with a missed conditional logic for route leave when the user is logging out. These were not included in the previous bug fix for tracking changes in drafts.

**Related issue(s)**:
#1547


jbair06 (approved):
LGTM

PR #1556, opened 2025-02-17
feat: add auto mnemonic matching

### LINKED ISSUE #1245: Auto mnemonic key matching
A user should be provided with an option to match 'imported' keys to a mnemonic. This flow would request the mnemonic from the user and would then try to match any key that has no mnemonic hash to the provided mnemonic. The user should also be given a limit to how many indices should be attempted for each key. This would enable users to attempt to match any unmatched key to the first x number of indices from a mnemonic. 

Once a key is matched, the hash of the mnemonic and the found index can be added to the key.

SvetBorislavov:
**Description**:
This pull request creates a page for matching recovery phrases to an external key

**Related issue(s)**:

Fixes #1245

### DIFF
--- a/front-end/src/main/modules/ipcHandlers/localUser/keyPairs.ts
+++ b/front-end/src/main/modules/ipcHandlers/localUser/keyPairs.ts
@@ -11,6 +11,7 @@ import {
   deleteKeyPair,
   updateNickname,
   updateMnemonicHash,
+  updateIndex,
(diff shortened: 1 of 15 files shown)


yiliev0 (dismissed):
LGTM

jbair06:
LGTM

jbair06 (approved):

Issue #1536, opened 2025-02-11
Autocomplete functionality should be ignored if the user presses escape


jbair06:
If the user is entering information into an autocomplete component and presses the 'escape' key, the component should not attempt to fill in the rest of the data. 

I have an account (0.0.258xxx) which I have the key for. If I wanted to set the fee payer to be account 0.0.2, the autocomplete feature prevents me from doing this. I tried to enter 0.0.2 then press enter, or tab. Both filled in the rest of the account (0.0.258xxx) as would be expected. When I pressed escape, the popup disappeared and the field only showed 0.0.2, but then I pressed tab or enter, and the field fills in the rest (0.0.258xxx). If I click out of the field at any time, the current actual type value (0.0.2) remains as expected.

The solution would be to ensure that when 'escape' is pressed, the autocomplete does not fill in the rest of the field when tab or enter is then pressed.

jbair06:
After pressing escape, the right arrow key shows the popup again, and adds the next digit as if the popup was never invisible. This is likely the expected behavior and should still remain after the solution is implemented for the described issue.

Issue #1534, opened 2025-02-11
Fee Payer Account ID autocomplete didn't scroll to account as expected


jbair06:
I have a list of more than 10 accounts that appear in my 'fee payer account id' drop down. When I start typing the account id of an account that is outside of the visible list, the drop down does not scroll to the item.

PR #1557, opened 2025-02-17
fix: scrolling to auto-suggested  in AppAutoComplete, fix: handler function

icoxxx:
**Description**:
This PR includes two fixes for the AppAutoComplete reusable component: ensuring correct scrolling to the auto-suggested account in the dropdown while typing and modifying the handler function logic for the keydown event.

**Related issue(s)**:
#1534 #1536


yiliev0 (approved):
LGTM, re-tested both issues and looks they are working as expected.

Issue #1376, opened 2024-12-17
Canceling a transaction in transaction group fails every other transaction


yiliev0:
### Description

Lets say we have 5 account create tx's in a group transaction. We initiate the group transaction, sign 4 of these and cancel the 5th. This causes all the other 4 transaction to be in "pending" state and expire after some time.

### Steps to reproduce

1. Create a user and sign into an organization
2. Crete a transaction group with a few transactions inside
3. Sign all transactions but keep one
4. For the last transaction click "cancel" 

### Version

0.5.0

### Operating system

None

jbair06:
Transactions in group should be individually cancel-able. If canceled, all apps should remove the canceled transactions from their view. The transactions should not be retrieved from future api requests. The canceled transactions should also not be processed.

PR #1558, opened 2025-02-17
feat: filter canceled transactions in a group

icoxxx:
**Description**:
Added a filter in the backend's execute service to exclude canceled transactions within a group. Since executeTransactionGroup validates all submitted transactions, this ensures that any transaction canceled in the meantime is properly filtered out.

**Related issue(s)**:
#1376

### DIFF
--- a/back-end/apps/chain/src/execute/execute.service.spec.ts
+++ b/back-end/apps/chain/src/execute/execute.service.spec.ts
@@ -576,5 +576,62 @@ describe('ExecuteService', () => {
         `Transaction Group cannot be submitted. Error validating transaction 0: ${errorMessage}`,
       );
     });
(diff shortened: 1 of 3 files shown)


SvetBorislavov (approved):
@jbair06  It looks alright, if we decide to ignore the 'atomic' flag

jbair06:
When we add the 'atomic' feature, then yes we can ignore the 'atomic' flag for canceled transactions.

jbair06:
@yiliev0 yes, it should work as you explained. If 9 of the 10 transactions are signed, and 1 is expired, the 9 should still be submitted and not fail.

yiliev0 (approved):
LGTM

PR #1567, opened 2025-02-20
Export Transaction as .txsig

jbair06:
**Description**:
Patch 0.9.1 altered how exporting worked. The exported transaction will now have the extension of .txsig. 

This was done to ensure other work flows that use these exported files don't need to change how they work. Also, the concept behind .tx and .txsig was that .tx only contains the transaction, where as .txsig also contains the signatures. All exported transactions from this app will contain signatures, if applicable. 

**Checklist**

- [X] Tested (unit, integration, etc.)

### DIFF
--- a/front-end/electron-builder.yml
+++ b/front-end/electron-builder.yml
@@ -22,10 +22,10 @@ nsis:
 mac:
   notarize: false
   target:
-    - target: 'default'
-      arch:
-        - x64
-        - arm64
+#    - target: 'default'
+#      arch:
+#        - x64
+#        - arm64
     - target: 'pkg'
       arch:
         - x64
--- a/front-end/src/main/modules/ipcHandlers/utils.ts
+++ b/front-end/src/main/modules/ipcHandlers/utils.ts
@@ -1,5 +1,4 @@
 import fs from 'fs';
-import path from 'path';
 
 import { BrowserWindow, app, dialog, ipcMain, shell, FileFilter } from 'electron';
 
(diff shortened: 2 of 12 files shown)


front-end/electron-builder.yml
```diff
@@ -22,10 +22,10 @@ nsis:
 mac:
   notarize: false
   target:
-    - target: 'default'
-      arch:
-        - x64
-        - arm64
+#    - target: 'default'
```
SvetBorislavov, line 25: Please don't comment the default target

front-end/src/main/modules/ipcHandlers/utils.ts
```diff
@@ -68,35 +67,27 @@ export default () => {
...
+          filters,
+          message,
+        });
+
+        if (canceled) return;
+        if (!filePath) throw new Error('File path is undefined');
+
+        await fs.promises.writeFile(filePath, data);
+      } catch (error: any) {
+        dialog.showErrorBox('Failed to save file', error?.message || 'Unknown error');
```
SvetBorislavov, line 89: This line is not covered, also don't use type any. It should be unknown and then validate if the type is error

front-end/src/main/modules/ipcHandlers/utils.ts
```diff
@@ -123,11 +114,11 @@ export default () => {
     },
   );
 
-  ipcMain.handle(createChannelName('sha384'), async (_e, str: string): Promise<string> => {
-    return await createHash('sha384').update(str).digest('hex');
+  ipcMain.handle(createChannelName('sha384'), (_e, str: string): Promise<string> => {
```
SvetBorislavov, line 117: Do not remove the async, await. Odd behaviour occur if not awaited

front-end/src/main/modules/ipcHandlers/utils.ts
```diff
@@ -123,11 +114,11 @@ export default () => {
...
+    return createHash('sha384').update(str).digest('hex');
   });
 
-  ipcMain.handle(createChannelName('x509BytesFromPem'), async (_e, pem: string | Uint8Array) => {
```
SvetBorislavov, line 130: Here also

front-end/src/renderer/services/electronUtilsService.ts
```diff
@@ -97,7 +95,7 @@ export async function x509BytesFromPem(pem: string | Uint8Array) {
 }
 
 /* Quits the application */
-export const quit = async () =>
```
SvetBorislavov, line 100: Here also don't remove the async/await

SvetBorislavov (changes requested):

front-end/electron-builder.yml
jbair06: Whoops, thanks for catching this. We don't current build the 'default' target. I commented it out to build the artifacts we needed. Didn't mean to upload this. I'll fix it.

SvetBorislavov (approved):

PR #1593, opened 2025-03-04
feat: verify node keys in DAB transactions

SvetBorislavov:
**Description**:
This pull request adjusts the requirements for the DAB transactions to the latest signing requirements

### DIFF
--- a/back-end/libs/common/src/utils/sdk/transaction.ts
+++ b/back-end/libs/common/src/utils/sdk/transaction.ts
@@ -18,6 +18,7 @@ import {
   parseAccountInfo,
   parseNodeInfo,
   safeAwait,
+  COUNCIL_ACCOUNTS,
 } from '@app/common';
 
 export const isExpired = (transaction: SDKTransaction) => {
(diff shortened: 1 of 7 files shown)


back-end/libs/common/src/utils/sdk/transaction.ts
```diff
@@ -200,10 +204,24 @@ export const computeSignatureKey = async (
...
           );
           if (accountInfo?.key) {
-            signatureKey.push(accountInfo.key);
+            nodeSignatureKey.push(accountInfo.key);
+          }
+        }
+      }
+
+      if (transactionIs(NodeDeleteTransaction, transaction)) {
+        const COUNCIL_ACCOUNTS = ['0.0.2', '0.0.50', '0.0.55'];
```
jbair06, line 213: Is there a better place to store this const to be reused throughout the app? Instead of duplicating it.

jbair06 (changes requested):

back-end/libs/common/src/utils/sdk/transaction.ts
jbair06: Just in case it were for some reason to change in the future (accounts added or removed)

jbair06 (approved):

PR #1614, opened 2025-03-17
Fix vulnerable dependencies

### LINKED ISSUE #1613: Fix dependency vulnerabilities
There are two critical vulnerabilities, which dependabot failed to resolve and must be resolved as soon as possible

SvetBorislavov:
**Description**:
This pull request updates vulnerable packages.
!Important note: The @hashgraph/sdk has a newer version, in which the SDK is minified. This introduces specific breaking changes. The current PR does not update the SDK to the minified version as updates from the SDK team should happen that will smoothen the update process here

**Related issue(s)**:

Fixes #1613

### DIFF
--- a/front-end/src/renderer/utils/index.ts
+++ b/front-end/src/renderer/utils/index.ts
@@ -188,10 +188,12 @@ export const getAccountIdWithChecksum = (accountId: string): string => {
 };
 
 export function stringifyHbarWithFont(hbar: Hbar, fontClass: string): string {
-  const tinybars = Math.abs(hbar.toTinybars());
-  const isHbar = tinybars >= Hbar.fromTinybars(1_000_000).toTinybars();
+  const tinybars = Math.abs(hbar.toTinybars().toNumber());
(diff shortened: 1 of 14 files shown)


front-end/src/renderer/utils/index.ts
```diff
@@ -188,10 +188,12 @@ export const getAccountIdWithChecksum = (accountId: string): string => {
 };
 
 export function stringifyHbarWithFont(hbar: Hbar, fontClass: string): string {
-  const tinybars = Math.abs(hbar.toTinybars());
-  const isHbar = tinybars >= Hbar.fromTinybars(1_000_000).toTinybars();
+  const tinybars = Math.abs(hbar.toTinybars().toNumber());
```
jbair06, line 191: tiny bars should already be a long. How does toNumber help here?

SvetBorislavov: There is a typecheck problem with that

SvetBorislavov: Fixed

PR #1619, opened 2025-03-19
feat: add public key setting option in contact list

### LINKED ISSUE #1575: Public key nicknaming in contact list and public key list
When viewing a Contact in the Contacts page, there is a list of public keys owned by that contact. This list shows 'Public Key' for each public key. With the adding of nicknames to public keys, this page should be altered and adapted to the new nicknaming feature for public keys.

The user should be able to change the nickname of a public key belonging to any contact. If the user changes the nickname of their own key, this should be reflected in the 'Settings' -> 'Key' page. If the user changes the nickname of a key belonging to another user, it should add that nickname to the 'Settings' -> 'Public Key' page along with the public key. 

How the UX should work for these alterations still needs to be determined. The following is one concept:

The public key list should be a table. Then each row would have the nickname of the public key (if applicable) and the public key. If the table row can be expanded dynamically to allow for the 'associated accounts' feature to remain, then that is preferable. If not, then a button can be added to the row that will allow the user to view the associated accounts. The nickname column would be editable.

SvetBorislavov:
**Description**:
This pull request displays the nickname of the public key in the contact list, while allowing it to be editable

**Related issue(s)**:

Fixes #1575


jbair06 (approved):
Looks great

PR #1660, opened 2025-05-15
feat: Add docker image builds

beeradb:
This PR introduces a new github workflow which builds the chain, API, and notifications containers on every tag of this repository. 

Additionally, it allows you to do a workflow dispatch to build the main branch.

### DIFF
--- a/.github/workflows/publish-docker-images.yaml
+++ b/.github/workflows/publish-docker-images.yaml
@@ -34,6 +34,7 @@ jobs:
           registry: ${{ env.REGISTRY }}
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
+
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-docker-images.yaml
```diff
@@ -0,0 +1,58 @@
...
+  REGISTRY: ghcr.io
+
+jobs:
+  build-and-push-image:
+    runs-on: transaction-tools-linux-medium
+    permissions:
+      contents: read
+      packages: write
+      attestations: write
+      id-token: write
```
rbarker-dev, line 19: I would move this to the top-level definition. But definitely not a blocker

rbarker-dev (changes requested):

rbarker-dev (approved):
LGTM

PR #1678, opened 2025-06-04
Helm chart

shezaan-hashgraph:
- This is a helm chart for hedera-transaction-tool. It was requested since SOLO only works with Helm charts.
- This chart has been tested locally using KinD for both functionality and template rendering.
- Eventually, the maintenance of this chart will be the responsibility of the Release Engineering or SOLO team - TBD
- This chart is self-contained and allows for local deployment to a `kind` cluster or `minikube` depending on your `values.yaml` file.


jbair06 (approved):
Looks good

PR #1696, opened 2025-06-18
Init stackgres

shezaan-hashgraph:
Use stackgres operator instead of cnpg for DB.


charts/transaction-tool/README.md
```diff
@@ -0,0 +1,7 @@
+If you want to create a simple postgres single pod cluster for dev purposes simply set `stackgres.enabled` to `false`
```
beeradb, line 1: I believe this should support 3 modes:

- No DB
- Small Dev DB
- "Production" DB

charts/transaction-tool/templates/secrets.yaml
```diff
@@ -41,41 +15,39 @@ metadata:
...
-  name: {{ .Release.Namespace }}-db-superuser
-  namespace: {{ .Release.Namespace }}
-  annotations:
-    "helm.sh/resource-policy": keep
-{{- if .Values.global.annotations }}
-    {{- toYaml .Values.global.annotations | nindent 4 }}
+  dbusername: dHJhbnNhY3Rpb24tdG9vbA==
+  dbpassword: {{ $b64UserPassword }}
+  dbname: dHJhbnNhY3Rpb24tdG9vbA==
+  script.sql: |
```
beeradb, line 21: does this only run once, or every startup?

Just curious if it will cause errors on pod restarts or anything like that.

jbair06 (approved):

PR #1784, opened 2025-08-29
chore: replace Hedera Local Node with Solo

SimiHunjan:
**Description**: This will replace local node with Solo in CI to run tests 

**Related issue(s)**:

Fixes #

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -1452,7 +1452,7 @@ The Initial Developer of copy-anything@3.0.5,
 is Luca Ban - Mesqueeb (https://github.com/mesqueeb/copy-anything).
 Copyright Luca Ban - Mesqueeb. All Rights Reserved.
 
-The Initial Developer of core-util-is@1.0.2,
+The Initial Developer of core-util-is@1.0.3,
 is Isaac Z. Schlueter (https://github.com/isaacs/core-util-is).
 Copyright Isaac Z. Schlueter. All Rights Reserved.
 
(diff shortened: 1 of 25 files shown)


jbair06 (dismissed):
Looks like solo will make the tests take a bit longer. But this is the right thing to do. Nice work! Anyone using Solo for local testing will need to be sure to use 0.43.0+, I think.

.github/workflows/test-frontend.yaml
```diff
@@ -102,15 +102,45 @@ jobs:
...
+          node_image: kindest/node:v1.31.4@sha256:2cb39f7295fe7eafee0842b1052a599a4fb0f8bcf3f83d96c7f4864c357c6c30
+          version: v0.26.0
+          kubectl_version: v1.31.4
+          verbosity: 3
+          wait: 120s
+
+      - name: Install Solo CLI
+        run: |
+          set -euo pipefail
+          npm install -g @hashgraph/solo
```
rbarker-dev, line 122:
```suggestion
          npm install -g @hashgraph/solo@${{ vars.SOLO_VERSION }}
```

We should set solo to a specific version. In consensus node we use the repository variable `CITR_SOLO_VERSION` for our testing pipeline and that gets updated periodically with solo releases. I would recommend having something similar here.

Note `${{ vars.SOLO_VERSION }}` would need to be defined for this repository.

rbarker-dev (approved):
1 nit.

.github/workflows/test-frontend.yaml
SimiHunjan: created a followup issue here for pinning Solo versions: https://github.com/hashgraph/hedera-transaction-tool/issues/1866

PR #1806, opened 2025-09-09
fix: do not ask to save new transactions without any alteration

### LINKED ISSUE #1794: New Transaction 'Back' button functionality
If the user attempts to create a new transaction but never makes any alterations to the transaction itself, the 'Back' button should not ask the user to save changes. 

Similar to https://github.com/hashgraph/hedera-transaction-tool/pull/1787

svienot:
**Description**:

Track various possible user edits to be able to assess that a new transaction has not been touched by the user, and hence does not require to display the Discard/Save modal dialog.

**Related issue(s)**:

Fixes #1794

**Notes for reviewer**:


svienot:
This approach is currently fully instantiated only for CreateAccount tx. Waiting for further feedback on the approach before moving forward, since this seems a bit convoluted compared to the benefit.

PR #1813, opened 2025-09-16
feat: add individual sign button to transaction group items

### LINKED ISSUE #1803: Add a "Sign" button to the group transactions next to "Details"
### Problem

Add a "Sign" button next to the "Details" button for group transactions. Expand some more of the transaction details in the row of the transaction task.

### Solution

see above

### Alternatives

_No response_

svienot:
**Description**:

Modify the Transaction Details page to:
 - allow to sign each group item individually
 - add an icon to reflect transaction status of group items

**Related issue(s)**:

Fixes #1803

**Notes for reviewer**:

Screenshot 2025-09-29 at 23 28 17

Screenshot 2025-09-29 at 23 24 32
Screenshot 2025-09-29 at 23 27 02
Screenshot 2025-09-29 at 23 25 18
Screenshot 2025-09-29 at 23 26 02


svienot:
Not sure how to reproduce this issue with the tooltips..

jbair06:
To recap our discussion this morning:
The tooltip value and visibility state do not appear to be reactive. It doesn't change when the transaction state changes. If no icon exists, then no tooltip is present, but when the icon appears, the tooltip does not appear.

However, it appears that the Mac's tooltip stuff overrides it. In the images above, that is what I seem to be seeing. The tooltip that doesn't look to be formatted seems to be cause by Mac, or the main process, or something. The reason I think this is because the tooltip was not visible when I was sharing my screen (like any other modal or popup that is Mac based, outside of the renderer's scope). 

Here is another image of the issue. In this case, the transaction had a status (I signed it, left the page and came back). Then, after the status changed, due to execution, the formatted tooltip remained unchanged, but the other tooltip appeared, and had the correct text: 
Screenshot 2025-09-30 at 8 43 56 AM

svienot:
Thanks for catching this and taking the time to explain the issue so clearly.
In short: once initialised, the bootstrap tooltips cannot be updated. 
So in this context, the only way is to recreate them at each update of the group items statuses (i.e. in handleFetchGroup())

PR #1814, opened 2025-09-16
fix: do not ask to save new unaltered transaction drafts

### LINKED ISSUE #1794: New Transaction 'Back' button functionality

svienot:
**Description**:

When leaving draft editing, we compare the transaction data (excluding the validStart) with a snapshot taken before the edit to determine whether we need to show the discard/save modal.

When creating a draft from scratch, we set a 500ms timer after the view is mounted to make sure the transaction is fully initialized before we can snapshot it.

**Related issue(s)**:

Fixes #1794

**Notes for reviewer**:

This obsoletes the draft PR #1806, which I am going to close.

### DIFF
--- a/front-end/src/renderer/components/Transaction/Create/BaseTransaction/BaseTransaction.vue
+++ b/front-end/src/renderer/components/Transaction/Create/BaseTransaction/BaseTransaction.vue
@@ -1,15 +1,21 @@
 <script setup lang="ts">
 import type { TransactionApproverDto } from '@shared/interfaces/organization/approvers';
-import { transactionsDataMatch, type TransactionCommonData } from '@renderer/utils/sdk';
 import {
+  getTransactionCommonData,
+  type TransactionCommonData,
+  transactionsDataMatch,
+  validate100CharInput,
+} from '@renderer/utils/sdk';
+import TransactionProcessor, {
   CustomRequest,
   type ExecutedData,
   type ExecutedSuccessData,
+  TransactionRequest,
 } from '@renderer/components/Transaction/TransactionProcessor';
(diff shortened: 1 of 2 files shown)


jbair06 (changes requested):
It appears that when I go into a draft, manually change the valid start to a future time, then 'update draft', it does not save the new value. Likewise, if I change the valid start to a future time and attempt to navigate away, it doesn't see that as a change. 

In the first case, it should save the change even if the valid start was the only thing to change (even if the user didn't manually change it, and it was changed by the ticking clock). 

In the second case, if the user changes the valid start to a future time, it should prompt the user to save. But it should not prompt the user to save if valid start is the only thing to change AND its value is <= now.

svienot:
Unless I am missing something, I cannot reproduce the behaviour you describe.
I just tested again that branch (after merging with main). Here is the behaviour I observe:
 - create new tx / enter a **future** date for valid start / click back
 - create new tx / enter a **past** date for valid start / UI automatically updates to current date and starts incrementing / click back
   --> Save modal dialog is NOT shown
 - create new tx / enter a **future** date for valid start / modify it to a **past** date / UI updates to now.... etc / click back
   --> Save modal dialog is NOT shown

jbair06:
To log our call: I must not have had the latest code pulled as I thought I had. I do not see the issues with draft any longer. If you also don't see any more issues, I'll approve.

jbair06 (approved):
Looks good!

PR #1843, opened 2025-09-25
feat: export transaction groups to v1 format

### LINKED ISSUE #1754: TTv2 exporting transaction groups
Similar to other exports, transaction groups must be exportable. ~~For TTv1 format, this means using the csva file type.~~ CSVA cannot be supported due to Java and JS SDK differences.

svienot:
**Description**:

Add an `Export` button the Transaction Group Details page which produces a zip file containing for each group item the `.tx` and `.txt` files identical to the files otherwise obtained by using `Export` on a single transaction.

Also a bit of refactoring of the single transaction (v1 format) export logic in order to share code.

**Related issue(s)**:

Fixes #1754


svienot:
These is one difference in the `.txt` file compared to the transaction export: the `author` value is not present, because the `creatorEmail` property I was expecting to find in the `ITransaction` is not defined (`item.transaction.creatorEmail`) in line 344 of TransactionGroupDetails.vue.

jbair06:
When I exported a transaction group I noticed that the .txt file created for TTv1 format didn't have content. It should have the content be equal to the transaction's description. If it has no description, it should fall back to the transaction group's description.

This image shows the transaction group description, and the content had no value in the .txt. This transaction also had its own description.
Screenshot 2025-09-30 at 12 28 46 PM

This image shows the content being filled when exporting a single transaction.
Screenshot 2025-09-30 at 12 28 53 PM

svienot:
I confirm the empty `content` is due to https://github.com/hashgraph/hedera-transaction-tool/issues/1857 (description is not stored in the DB when the tx is a group item). 
I have changed the implementation such that when the tx description is not present, the content defaults to the tx group description.

jbair06 (approved):
Look good!

PR #1860, opened 2025-10-02
Added missing return at the end of getServiceEndpoint()

ericleponner:
**Description**:

`createTransaction.getServiceEndpoint()` is missing a final `return null`.
This makes its return type to be `ServiceEndpoint | null | undefined` (in place of` ServiceEndpoint | null`) and `mirrorNodeDataService.getNodeInfo()` type checking unhappy.
Changes below add the missing `return null`.

**Checklist**


svienot (approved):
LG

PR #1882, opened 2025-10-09
fix: transaction group management make many duplicated calls to `api/v1/accounts/{accountId}

### LINKED ISSUE #1881: Transaction group management makes many duplicated calls to api/v1/accounts/{accountId}
### Description

Logic for displaying and signing transaction groups make a lof of use of api/v1/accounts/{accountId} call to mirror node. Many of them are duplicated and could be avoid with a transient cache.
This is typically true when:
- selecting In Progress tab
- running Sign All action
In many cases, duplicated calls are cached by Chromium and do not reach mirror node.
But still they trigger disk io and cpu load.

### Steps to reproduce

Calls to api/v1/accounts/{accountId} are visible in Dev Tools Console.

### Additional context

_No response_

### Hedera network

_No response_

### Version

0.61.1

### Operating system

macOS

ericleponner:
**Description**:

Changes below introduce new class `AccountInfoCache`. This encapsulates call to `api/v1/accounts/{accountId}` and caches returned results.

The following methods now run `AccountInfoCache.fetch()` in place of calling mirror node directly:
- `TransferHbar.handleDraftLoaded()`
- `TransferDetails`
- `MultipleAccountUpdateRequest.deriveRequestKey()`
- `CreateTransactionGroup.handleOnFileChange()`
- `TransactionBaseModel.computeSignatureKey()`

`AccountInfoCache` lifetime is limited to the scope of these methods.
This reduces effect of caching but avoids cache desynchronization issues.

**Related issue(s)**:

Fixes #1881

**Checklist**

### DIFF
--- a/front-end/src/renderer/components/Transaction/Create/BaseTransaction/BaseTransaction.vue
+++ b/front-end/src/renderer/components/Transaction/Create/BaseTransaction/BaseTransaction.vue
@@ -251,7 +251,11 @@ function basePreCreateAssert() {
 }
 
 async function updateTransactionKey() {
-  const computedKeys = await computeSignatureKey(transaction.value, network.mirrorNodeBaseURL, new AccountInfoCache());
+  const computedKeys = await computeSignatureKey(
+    transaction.value,
+    network.mirrorNodeBaseURL,
+    new AccountInfoCache(),
+  );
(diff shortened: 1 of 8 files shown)


svienot (dismissed):
LGTM

PR #1923, opened 2025-10-23
V0.18.1 quorum queues fix

jbair06:
**Description**:
RabbitMQ was using classic queues. This was potentially the cause of the intermittent communication issue we were seeing. The queues are now quorum queues which work better with kubernetes.


svienot (approved):
LGTM

PR #1933, opened 2025-10-27
chore: fix transaction group tests and add test for cancel all action

svienot:
Fix regression caused by recent changes to the `Sign All` button.
Add test for the new `Cancel All` feature.
Note: these tests are still unstable. They should pass individually but fail to pass when the whole test file is executed.

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -988,7 +988,7 @@ The Initial Developer of @vue/tsconfig@0.7.0,
 is Haoqun Jiang (https://github.com/vuejs/tsconfig).
 Copyright Haoqun Jiang. All Rights Reserved.
 
-The Initial Developer of @vuepic/vue-datepicker@10.0.0,
+The Initial Developer of @vuepic/vue-datepicker@11.0.2,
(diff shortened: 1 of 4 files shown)


automation/tests/organizationGroupTests.test.js
```diff
@@ -146,31 +147,44 @@ test.describe('Organization Group Tx tests', () => {
...
+    await groupPage.clickOnConfirmGroupTransactionButton();
+    await groupPage.clickOnCancelAllButton();
+    await groupPage.clickOnConfirmGroupActionButton()
+    await loginPage.waitForToastToDisappear();
+    await transactionPage.clickOnTransactionsMenuButton();
+    await organizationPage.clickOnReadyToSignTab()
+    expect(await groupPage.isEmptyTransactionTextVisible()).toBe(true);
+  });
+
+  test(`Verify user can import csv with 100 transactions`, async () => {
```
jbair06: This test was set to run with 5 and 100 because it was going through all the signing of the transactions using the 2 different flows: sign all and sign each individually. Do we still test the 'sign individually' flow?

PR #1949, opened 2025-10-30
1785 back end transaction group execution speed issue

### LINKED ISSUE #1785: Back-end transaction group execution speed issue
Currently, a transaction group will be sent out from a single pod. If the transaction group is NOT sequential, that means that one pod is trying to send out all the transactions for the group simultaneously. This can become very resource intensive. 

One reason for this design was to ensure that the group is only executed if all transactions are valid. There was an atomic flag that was added, partially for this purpose, but may not be in current use. 

Assuming 'atomic' is not used, or if it is and atomic is false, then the pod that picks up the transaction group should determine its sequential status, then process them together if sequential, or allow other pods to pick up transactions from the group.

jbair06:
**Description**:
In most cases, a transaction group does not need to be processed together. Unless 'atomic' (which is not yet supported) or 'sequential', there is no expectation from the user that all transactions will be run at the same time and nor that they will all be validated before any are executed. Now, transactions that are part of a group that is neither atomic nor sequential will be processed as any other transaction. This will lower the processing power required to handle large groups. 

**Related issue(s)**:

Fixes #1785 

**Notes for reviewer**:
Further improvements can be made, but will require a fair amount of work in order to properly spread out all the work among the different chain services replicas without causing bottle necks and backlogs.

**Checklist**

### DIFF
--- a/automation/tests/organizationTransactionTests.test.js
+++ b/automation/tests/organizationTransactionTests.test.js
@@ -231,7 +231,7 @@ test.describe('Organization Transaction tests', () => {
     const { txId, validStart } = await organizationPage.updateAccount(
       complexKeyAccountId,
       'update',
-      3000,
+      10000,
       true,
     );
     await transactionPage.clickOnTransactionsMenuButton();


svienot (approved):
LGTM

ericleponner (approved):
Looks good.

PR #1952, opened 2025-10-30
fix: user repeatedly asked to enter mnemonic

### LINKED ISSUE #1942: User repeatedly asked to enter mnemonic on app startup and 'Personal' mode selection
In a previous version of the app, only a single mnemonic was supported. Once multi mnemonic support was added, a connection was needed from the previous created keys to the mnemonic used to create the key. This was done by introducing a modal at launch, requiring the user's previously entered mnemonic. The issue is this modal, if skipped, will return the next time it is used.

The Mnemonic Migration Modal needs to mark keys as not requiring a mnemonic, if the modal is skipped.

svienot:
**Description**:

When the user chooses to skip the step where "_previously created private keys need to be rematched to a mnemonic_" we make sure this step is never presented again (for that mode) by resetting the `secret_hash` and `index` fields of the corresponding _KeyPairs_ in the local DB.

From that point, any such key pair will show up as _Imported from Private Key_ under the _Settings > Keys_ tab. If the user wishes to perform the rematching, they can select the _Match Recovery Phrase_ action.

**Related issue(s)**:

Fixes #1942

** Notes for reviewer**:

One simple way to artificially trigger the display of the _rematching_ step is to set the `secret_hash` property of the _KeyPair_ to a random string in the local DB.

### DIFF
--- a/front-end/src/renderer/components/Transaction/Create/AccountDelete/AccountDelete.vue
+++ b/front-end/src/renderer/components/Transaction/Create/AccountDelete/AccountDelete.vue
@@ -4,7 +4,7 @@ import type { CreateTransactionFunc } from '@renderer/components/Transaction/Cre
 import type { AccountDeleteData } from '@renderer/utils/sdk';
 
 import { computed, onMounted, reactive, ref, watch } from 'vue';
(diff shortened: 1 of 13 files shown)


ericleponner (approved):
LGTM

PR #1992, opened 2025-11-07
fix: add watching logic to storeContacts and avoid unneeded calls to storeContacts.fetch()

ericleponner:
**Description**:

Changes below add watching logic to `storeContacts`.
They avoid some unneeded calls to `storeContacts.fetch()` at application startup.

**Checklist**


svienot (approved):
LGTM

PR #1995, opened 2025-11-07
fix: moving router.push() in AccountSetup.handleNext()

ericleponner:
**Description**:

`AccountSetup.handleNext()` is the method called each time user clicks on `Next` button in `Account Setup` wizard.
When last wizard step is reached, `handleNext()` invoke `router.push('/transactions')`.
This invocation is done at the beginning of `handleNext()` without await operator.
In some cases, it causes key saving to be executed two times (and two` Key Pair saved successfully` toasts displayed).
Changes below move `router.push()` at the end of `handleNext()` execution.

They also loosen `registrationTests.settingsButtonSelector` by removing `active` class.
Indeed `AccountSetup.handleNext()` re-route to `/transactions` and in this case `Settings` button is never active.

**Checklist**


svienot (approved):
LGTM

PR #2026, opened 2025-11-25
feat: allow to sign transaction groups from the Ready to Sign table.

svienot:
**Description**:

Provide a `Sign All` button in the `Ready to Sign` table when the row item is a Transaction Group (note this action brings up a confirmation dialog).

Also replace the `signSingleTransaction()` function by `signTransactions()` in renderer utils and make use of that in various places where there is a Sign action.

**Notes for reviewer**:

Screenshot 2025-11-25 at 22 41 37

Screenshot 2025-11-25 at 22 41 32


ericleponner (approved):
LGTM

Issue #1906, opened 2025-10-16
Manual 'Sign Transaction' reminder notification does not clear from notification popup


jbair06:
I created a transaction, then shared it with the org. I then manually clicked 'Remind Signers'. Now that signer has a notification in the notification popup that won't go away even when clicked. The database shows it to be like any other notification, but isRead is false, and they type is 'TRANSACTION_WAITING_FOR_SIGNATURES_REMINDER_MANUAL'. 

The notification should appear the same as if it were the original notification. It should replace the original notification, if it still exists, so both notifications don't show up in the app. Once the notification is clicked (or the transaction is viewed) the notification should be removed, just like the original notification sent to signers.

<img width="447" height="135" alt="Image" src="https://github.com/user-attachments/assets/f12200a7-0157-417b-8d9d-7f9f9fd73f02" />

Issue #1965, opened 2025-10-31
Uploading signatures creates too many transaction_signer entities


jbair06:
When a user signs a transaction, the signature gets uploaded to the backend. The signature is actually multiple signatures, one for each node in the list of nodes in the transaction. 

The signature bytes are not stored in the database inside the transaction_signer table. It is stored in the transaction itself. For this reason, there is no need for multiple transaction_signer entities.

Be sure that uploading signatures only saves a single transaction_signer entity per userKey per transaction.

Issue #1929, opened 2025-10-24
Refactor Backend Notification System


jbair06:
The notification system, as implemented, on the backend is not actually scalable. Currently, all messaging sent from one service to another is fire-and-forget, and is only processed by a single replica. This means that when api sends anything to be processed by chain, RabbitMQ only allows 1 chain service to get the message and process it. Same thing with the notification service.

**Issues with design**
With the current design, there is no acknowledgement that the consumer of the message actually consumes the message. In addition, in app notifications need to be fanned out to all connected users. It works for now due to all backend deployments using a single notification replica. If more are provisioned, it will no longer work.

**Solution**
All consumers should acknowledge, always. Whether manually acknowledged (for EventPattern where RabbitMQ needs to know), or automatically (for MessagePattern where the producer is waiting), we don't want any messages to get lost.  
Email and InApp notifications will need to be separated into different queues. Email can be a simple queue, like it currently is. But InApp notifications needs to be a queue coupled with a fanout exchange. This means that a lot of the notification service code will need to be separated out, and the notification logic will need to be redesigned to handle these notifications separately.

Issue #1887, opened 2025-10-13
Cancel button still visible and active after cancelling a transaction


ericleponner:
### Description

When a transaction is ready for execution, `Transaction Details` displays a `Cancel` button.
Clicking this button cancels the transaction but button remains visible and active.
Clicking a second time displays `Only transactions in progress can be canceled` message.
`Cancel` button should be hidden once cancel action is completed.

### Steps to reproduce

1) Starts TT
2) Goto `Transactions` tab
3)` Create New` > `Transaction` > `Create Account`
4) `Payer ID` = me
4) `Valid Start` = far
5) `Create and Share`
6) `Sign`
7) Goto `Transaction` > `Ready for Execution` tab
   => newly created transaction appears
8) Open `Transaction Details`
9) `Cancel` + `Confirm`
   => Transaction is cancelled successfully
   => `Cancel` button is still visible and active
10) `Cancel` + `Confirm` again
   => `Only transactions in progress can be canceled` message is displayed

### Version

0.17.0

### Operating system

None

jbair06:
I will investigate if this is related to the websocket issues previously mentioned.

PR #2050, opened 2025-12-04
Messaging layer replacement and improvements

### LINKED ISSUE #1740: Potential intermittent WebSocket connection issue
I created an AccountUpdateTransaction. I sent the transaction to the back-end via 'Create and Share'. I then watched the transaction (via TransactionDetails page). The transaction expired, I saw the notification bell ring (which may or may not have been related to the expiring transaction) but the TransactionDetails page never updated. 

I tried a lot of different setups with AccountUpdateTransaction to get something consistent. I was unable to always reproduce this behavior. Though, once the connection appears to have broken, it remains so even for other notification types. 

This was tested with solo as the network, the backend being deployed locally via kubernetes.

jbair06:
**Description**:
RabbitMQ has some limitations due to our desire to maintain a lightweight backend. In order to properly set up RabbitMQ for our messaging needs, it would require a lot more resources. Nats + Jetstream will be used as the replacement. 

With this change, we now will be using a proper at-most-once queue for all messaging that should only be processed by a single replica of a single service. We will also be using a at-least-once fan out type queue for all broadcast communications to enable all connected websockets to push the proper notifications. 

Speed improvements were added as well. Transaction Group signing was a big issue. If the group was too large, and multiple keys were involved, the backend could not keep up. This causes a number of issues, including unresponsive UI, numerous emails/inapp notifications, etc. Now signing is done in a batched manner, all calls to the database are grouped were possible. 

**Related issue(s)**:

Fixes #1740 #1800 #1887 #1625 #1770 #1929 #1965 #1906

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/back-end/k8s/dev/shell/deployments.sh
+++ b/back-end/k8s/dev/shell/deployments.sh
@@ -45,13 +45,11 @@ assert_nats_release() {
         $HELM uninstall nats
     fi
 
+    local nats_values_yaml=$(realpath "$BASEDIR/deployments/nats-values.yaml")
+
     # In a normal install, we might want exporter and reloader
     # For a full dev, we want all of them, I think.
-    $HELM install nats nats/nats \
-            --set jetstream.enabled=true \
-            --set natsBox.enabled=false \
-            --set exporter.enabled=false \
-            --set reloader.enabled=false
+    $HELM install nats nats/nats -f "$nats_values_yaml"
 }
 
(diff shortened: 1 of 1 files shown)


jbair06:
I'm still checking to see if these fixes also resolve issues #1906 and #1625. Both should now be resolved, but I'm checking tests and manually verifying as well.

PR #2093, opened 2025-12-12
fix: transfer exceeds balance error

### LINKED ISSUE #1975: Transfer exceeds balance error issue
When the user enters an amount for a transfer transaction that is less than 0 tiny bar, the error is displayed. Furthermore, the error did not hide when I corrected the amount by removing a 0 from the amount.

Image

Noobushi:
Description:
Leading '.' are now converted to zero + '.' (. -> 0.)
There is no problem with the exceed amount message or behavior (0 <  0.00009).

Fixes #1975


Noobushi:
Taking into account the issue described above with the test setup, I’m closing this PR and will commit the fix directly to the repository instead.

PR #2098, opened 2025-12-15
chore: application launch improvements

### LINKED ISSUE #2073: Default to the last used organization for username and password
### Problem

Fix the flow when the user users username and password to create their profile to open to the last used organization.

### Solution

See above

### Alternatives

_No response_

### LINKED ISSUE #2074: Application launch improvements
### Problem

Avoid transient switch to personal mode, last organization is not remembered when the app is closed when the organization is first created. Investigate any other issues during application launch. 

- Review how the initialization is done

### Solution

See above

### Alternatives

_No response_

RAMTO:
**Description**:
This PR is improving the launch of the app, by setting the last created/used org
* set the last added org in the app state
* remove it from the state if the org is deleted

**Related issue(s)**:

Fixes #2073 
Fixes #2074 

**Notes for reviewer**:

**Checklist**

### DIFF
--- a/front-end/src/renderer/composables/user/useAfterOrganizationSelection.ts
+++ b/front-end/src/renderer/composables/user/useAfterOrganizationSelection.ts
@@ -5,6 +5,7 @@ import { useRouter } from 'vue-router';
 import { SKIPPED_PERSONAL_SETUP } from '@shared/constants';
 
 import useSetupStores from '@renderer/composables/user/useSetupStores';
+import useDefaultOrganization from '@renderer/composables/user/useDefaultOrganization';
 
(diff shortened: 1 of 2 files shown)


front-end/src/renderer/pages/OrganizationLogin/OrganizationLogin.vue
svienot: I am not 100% sure this change is needed. I think we need setLast() to be called:
 - when the user switches orgs (was already covered)
 - when an org is added (which you have covered)
 - when an org is deleted (which you have covered)

svienot: Integrating this since we need to move forward. The unneeded setLast() is harmless, we might remove it later.

PR #2116, opened 2025-12-16
Fix: button gap in transaction table

svienot:
**Description**:

Recent change in TransactionNodeRow to silence execution warnings related to data-testid has broken the buttons gap.


ericleponner (approved):
LGTM

Issue #2103, opened 2025-12-15
Enhance `version-check` to detect and return frontend updates


jbair06:
## Description

Extend the existing `version-check` endpoint so that, after logging and storing the user’s frontend version, the backend checks whether a newer supported frontend version is available based on environment variables. The endpoint will also expose a minimum supported frontend version so the client can decide when to require a mandatory upgrade. If a newer supported version exists, the endpoint returns an update URL to the frontend in the `version-check` response.

## Requirements

- After recording the user’s version, the endpoint should:
  - Read an environment variable containing the latest supported frontend version (e.g. `LATEST_SUPPORTED_FRONTEND_VERSION`).  
  - Read an environment variable containing the minimum supported frontend version (e.g. `MINIMUM_SUPPORTED_FRONTEND_VERSION`).
  - Read an environment variable containing the frontend repository or release URL base (e.g. `FRONTEND_REPO_URL`).  
  - Parse the user’s version from the request and compare it to `LATEST_SUPPORTED_FRONTEND_VERSION` and `MINIMUM_SUPPORTED_FRONTEND_VERSION` using semantic version comparison.  
- Version logic:
  - If `LATEST_SUPPORTED_FRONTEND_VERSION` is greater than the user’s current version, treat this as an available update.  
  - If the user’s version is less than `MINIMUM_SUPPORTED_FRONTEND_VERSION`, mark the client as below the minimum supported version.
  - If the user’s version is equal to or greater than `LATEST_SUPPORTED_FRONTEND_VERSION`, do not advertise an update.  
- Response behavior:
  - Always perform the original behavior: log and persist the user’s version.  
  - Include version metadata in the response payload, for example:  
    - `{ latestSupportedVersion: string, minimumSupportedVersion: string, updateUrl?: string }`.  
  - When an update is available (user version < latest supported), include an `updateUrl` derived from `FRONTEND_REPO_URL` and `LATEST_SUPPORTED_FRONTEND_VERSION`.  
  - When no newer supported version is available, omit `updateUrl` (or return it as `null`), but still return `latestSupportedVersion` and `minimumSupportedVersion` for transparency.   

## Acceptance Criteria

- `version-check` continues to log and persist the user’s version as before.  
- Response includes `latestSupportedVersion` and `minimumSupportedVersion` taken from environment variables.  
- When the user’s version is older than `LATEST_SUPPORTED_FRONTEND_VERSION`, the response includes an `updateUrl` and the latest supported version.  
[...]

PR #2119, opened 2025-12-17
feat: enhance version check to detect and return frontend updates

RAMTO:
**Description**:
Enhance the version log with additional checks for minimum and new available version

**Related issue(s)**:

Fixes https://github.com/hashgraph/hedera-transaction-tool/issues/2103

**Notes for reviewer**:
New .env variables are added in the `api` folder and in:
* `back-end/k8s/back-end/api-deployment.yaml`
* `back-end/k8s/dev/deployments/api-deployment.yaml`

**Checklist**

### DIFF
--- a/back-end/apps/api/src/users/dtos/version-check-response.dto.ts
+++ b/back-end/apps/api/src/users/dtos/version-check-response.dto.ts
@@ -17,17 +17,10 @@ export class VersionCheckResponseDto {
   minimumSupportedVersion: string;
 
   @ApiProperty({
-    description: 'URL to the releases page where the user can download updates',
+    description: 'URL to the releases page where the user can download updates. Null if no update is available.',
     example: 'https://github.com/hashgraph/hedera-transaction-tool/releases/tag/v1.2.0',
     nullable: true,
   })
   @Expose()
   updateUrl: string | null;
-
-  @ApiProperty({
-    description: 'True if the user version is older than the latest supported version',
-    example: true,
-  })
-  @Expose()
-  updateAvailable: boolean;
 }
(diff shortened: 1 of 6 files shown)


back-end/apps/api/src/users/dtos/version-check-response.dto.ts
```diff
@@ -8,4 +8,42 @@ export class VersionCheckResponseDto {
...
   @Expose()
   success: boolean;
+
+  @ApiProperty({
+    description: 'The latest supported frontend version (from server configuration)',
+    example: '1.2.0',
+    nullable: true,
+  })
+  @Expose()
+  latestSupportedVersion: string | null;
```
jbair06: Do the version strings need to be nullable? What does null mean other the frontend?

back-end/apps/api/src/users/dtos/version-check-response.dto.ts
```diff
@@ -8,4 +8,42 @@ export class VersionCheckResponseDto {
...
+  })
+  @Expose()
+  updateUrl: string | null;
+
+  @ApiProperty({
+    description: 'True if the user version is older than the latest supported version',
+    example: true,
+  })
+  @Expose()
+  updateAvailable: boolean;
```
jbair06: If updateUrl is null, then the updateAvailable should be false. Is there any other reason to have this property?

back-end/apps/api/src/users/users.controller.ts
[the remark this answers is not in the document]
RAMTO: `UserDto` returns the whole user object, in this case the function is returning a string, so the dto won't be necessary here.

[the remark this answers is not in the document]
RAMTO: `UserDto` returns the whole user object, in this case the function is returning a boolean, so the dto won't be necessary here.

back-end/apps/api/src/users/dtos/version-check-response.dto.ts
RAMTO: It won't be null, cause it is taken from the .env 
made it required, so it won't be null

back-end/apps/api/src/users/dtos/version-check-response.dto.ts
RAMTO: yes, you are right, the front end will check the `updateUrl` in order to decide if an update is needed

PR #2122, opened 2025-12-17
Feat: add conditional tooltip on hover in the transaction overview table

Noobushi:
Description:
The transaction overview table now shows a minimal hover tooltip with the full description text.
The tooltip is displayed only when the text exceeds the available space to avoid unnecessary overlap.

### DIFF
--- a/front-end/src/renderer/pages/Transactions/components/TransactionNodeRow.vue
+++ b/front-end/src/renderer/pages/Transactions/components/TransactionNodeRow.vue
@@ -172,7 +172,7 @@ onUnmounted(() => {
 /* Watchers */
 watch(() => props.node.description, () => {
   nextTick(() => checkTruncation());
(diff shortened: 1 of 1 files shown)


front-end/src/renderer/pages/Transactions/components/TransactionNodeRow.vue
```diff
@@ -125,6 +132,48 @@ const handleDetails = async () => {
...
+    checkTruncation();
+  });
+});
+
+onUnmounted(() => {
+  window.removeEventListener('resize', checkTruncation);
+});
+
+/* Watchers */
+watch(isTruncated, (newValue, oldValue) => {
```
jbair06: As isTruncated is only ever set in checkTruncation(), can we put this logic in checkTruncation and remove the watch? Or is there another reason for the watch?

PR #2155, opened 2026-01-09
fixed the select button text color issue in light/dark themes

ddeskov-limechain:
### Problem
The "Select" button in the Accounts and Files pages had poor contrast when selected. The button used a dark background (`bg-secondary` = `#333666`) but kept dark text (`text-dark-emphasis`), making the text unreadable.

### Solution
Made the text color conditional based on the selection state:
- **Unselected**: Dark text (`text-dark-emphasis`) on transparent background
- **Selected**: White text (`text-white`) on dark background (`bg-secondary`)

The icon retains the `text-headline` class for consistent sizing, and inherits the white color from the parent button when selected.

### Files Changed
- `front-end/src/renderer/pages/Accounts/Accounts.vue`
- `front-end/src/renderer/pages/Files/Files.vue`

## Before

image

## After

image

### DIFF
--- a/front-end/src/renderer/pages/Transactions/components/History.vue
+++ b/front-end/src/renderer/pages/Transactions/components/History.vue
@@ -31,8 +31,7 @@ import {
   isLoggedInOrganization,
   isUserLoggedIn,
 } from '@renderer/utils';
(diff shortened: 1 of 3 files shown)


ericleponner (approved):
LGTM

PR #2158, opened 2026-01-12
Showing the freeze type on the transaction list table

ddeskov-limechain:
**Problem**

To help move faster during releases and in the narrow time window for abort transactions, the Transaction `Freeze` type needs to be added in the` Transaction Type` column of the transaction list table.

**Solution**

The backend does not store the specific freeze transaction type (Freeze Only, Prepare Upgrade, Freeze Upgrade, Freeze Abort) - it only stores the generic "FREEZE" type in the database. The freeze subtype information is only available within the serialized transactionBytes.

To minimize the impact on the codebase and avoid backend changes, a **frontend-only approach** was implemented:

1. New utility function `getFreezeTypeForTransaction() `in **transactions.ts**:
    * When the frontend detects a freeze transaction, it makes an additional API call to `getTransactionById()` to fetch the full transaction details including transactionBytes
    * Deserializes the transaction bytes using the Hedera SDK
    * Extracts the `freezeType` code from the FreezeTransaction instance
    * Results are cached in a Map<number, number | null> to avoid redundant API calls for the same transaction
2. New utility function `getDisplayTransactionType() `in **transactions.ts**:
    * Accepts either an SDK Transaction or Uint8Array
    * For freeze transactions, extracts and returns the specific freeze type string
    * Falls back to standard transaction type for non-freeze transactions
3. Updated components:
    * **TransactionNodeRow.vue** (Ready to Sign, In Progress, Ready for Execution, History tabs in organization mode): Added a watcher that detects freeze transactions and fetches the freeze type on-demand
    * **History.vue** (personal mode): Uses the existing body field from local transactions to extract freeze type
    * **Drafts.vue**: Parses the transactionBytes field to display specific freeze types for draft transactions
    
 **Before**
 
Transaction ID

**After**
Transaction ID

### DIFF
--- a/front-end/src/renderer/utils/sdk/transactions.ts
+++ b/front-end/src/renderer/utils/sdk/transactions.ts
@@ -40,7 +40,7 @@ export const formatTransactionType = (
   }
   if (short) {
     // Remove all whitespace characters
-    result = type.replace(/\s+/g, '');
+    result = result.replace(/\s+/g, '');
   }
   return result;
 };
--- a/front-end/src/tests/renderer/utils/sdk/transactions.spec.ts
+++ b/front-end/src/tests/renderer/utils/sdk/transactions.spec.ts
@@ -1,0 +1,151 @@
+import { describe, test, expect, vi, beforeEach } from 'vitest';
+import { FreezeTransaction, FreezeType, TransferTransaction } from '@hashgraph/sdk';
+import {
+  getFreezeTypeString,
+  getDisplayTransactionType,
+  formatTransactionType,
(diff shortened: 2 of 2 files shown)


ddeskov-limechain:
I have tested it in Company and Personal mode, but please manually check it as well. I have just been added to the project and I am not aware of any corner cases, etc.

ddeskov-limechain:
Closed due to failing tests. Changes were moved to [PR-2170](https://github.com/hashgraph/hedera-transaction-tool/issues/2170)

PR #2159, opened 2026-01-12
Endpoint Input auto strips protocol

ddeskov-limechain:
**Problem**

When creating or updating a node, users may paste full URLs into the gossip, service, and gRPC web proxy endpoint fields. For example, a user might enter:

- http://my.domain.com
- https://my.domain.com/api/v1/
- my.domain.com/api/v1

These endpoint fields expect only a domain name or IP address, not a full URL with protocol or path. The previous implementation stored the input as-is, which would result in invalid endpoint configurations.

**Solution**

Added a `stripProtocolAndPath()` helper function in **NodeFormData.vue** that automatically sanitizes endpoint inputs by:

1. Removing protocol prefixes - Strips http://, https://, grpc://, and any other URI scheme prefixes
2. Removing URL paths - Strips everything after the first / (e.g., /api/v1/)
3. Preserving trailing dots - Keeps trailing . as it is valid in Fully Qualified Domain Names (FQDNs)

The helper is applied in two places:

`getEndpointData()` - Used when adding gossip and service endpoints
`getGrpcWebProxyEndpoint() `- Used when updating the gRPC web proxy domain

**Examples**

Input  | Expected result after adding
-- | --
http://my.domain.com | my.domain.com
https://my.domain.com/api/v1/ | my.domain.com
my.domain.com/api/v1 | my.domain.com
my.domain.com. | my.domain.com.
192.168.1.1 | 192.168.1.1

<br class="Apple-interchange-newline">

**Before**

image

**After**

image

### DIFF
--- a/front-end/src/renderer/components/Transaction/Create/NodeCreate/NodeFormData.vue
+++ b/front-end/src/renderer/components/Transaction/Create/NodeCreate/NodeFormData.vue
@@ -169,14 +169,32 @@ function handleInputValidation(e: Event) {
 }
 
 /* Functions */
+function stripProtocolAndPath(input: string): string {
+  let result = input.trim();
+
+  // Remove protocol prefix (http://, https://, grpc://, etc.)
+  result = result.replace(/^[a-zA-Z][a-zA-Z0-9+.-]*:\/\//, '');
+
+  // Remove path (everything after the first /)
+  // Preserve trailing dot for FQDN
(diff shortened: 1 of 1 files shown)


ddeskov-limechain:
Closing PR due to failing tests. Changes we moved to [PR-2171](https://github.com/hashgraph/hedera-transaction-tool/pull/2171)

Issue #1961, opened 2025-10-31
Select button uses incorrect colors when selected


jbair06:
The tabs like Accounts or Files, the 'select' button does not use the proper color scheme when selected. Perhaps it is just the text color that needs to be changed.

<img width="462" height="188" alt="Image" src="https://github.com/user-attachments/assets/b6eb843a-f25f-4654-9e77-e8c488582053" />

<img width="444" height="204" alt="Image" src="https://github.com/user-attachments/assets/92b65672-9326-4b58-b0c5-87e37479eb20" />

PR #2169, opened 2026-01-14
Fixed the select button text color issue in light/dark themes

ddeskov-limechain:
Fixes issue #1961

### Problem
The "Select" button in the Accounts and Files pages had poor contrast when selected. The button used a dark background (`bg-secondary` = `#333666`) but kept dark text (`text-dark-emphasis`), making the text unreadable.

### Solution
Made the text color conditional based on the selection state:
- **Unselected**: Dark text (`text-dark-emphasis`) on transparent background
- **Selected**: White text (`text-white`) on dark background (`bg-secondary`)

The icon retains the `text-headline` class for consistent sizing, and inherits the white color from the parent button when selected.

### Files Changed
- `front-end/src/renderer/pages/Accounts/Accounts.vue`
- `front-end/src/renderer/pages/Files/Files.vue`

## Before

image

## After

image


automation/pages/TransactionPage.ts
```diff
@@ -493,6 +493,8 @@ export class TransactionPage extends BasePage {
     await this.fillInDeletedAccountId(accountId);
     await this.clickOnSignAndSubmitButton();
     await this.clickOnConfirmDeleteAccountButton();
+    // Wait for delete confirmation modal to close before looking for transaction modal
+    await this.window.waitForTimeout(500);
```
jbair06: `this.clickSignTransactionButton()` already waits for the component to be visible. Does `this.window.waitForTimeout(500)` add anything extra to that?

PR #2170, opened 2026-01-14
Showing the freeze type on the transaction list table

ddeskov-limechain:
**Problem**

To help move faster during releases and in the narrow time window for abort transactions, the Transaction `Freeze` type needs to be added in the` Transaction Type` column of the transaction list table.

**Solution**

The backend does not store the specific freeze transaction type (Freeze Only, Prepare Upgrade, Freeze Upgrade, Freeze Abort) - it only stores the generic "FREEZE" type in the database. The freeze subtype information is only available within the serialized transactionBytes.

To minimize the impact on the codebase and avoid backend changes, a **frontend-only approach** was implemented:

1. New utility function `getFreezeTypeForTransaction() `in **transactions.ts**:
    * When the frontend detects a freeze transaction, it makes an additional API call to `getTransactionById()` to fetch the full transaction details including transactionBytes
    * Deserializes the transaction bytes using the Hedera SDK
    * Extracts the `freezeType` code from the FreezeTransaction instance
    * Results are cached in a Map<number, number | null> to avoid redundant API calls for the same transaction
2. New utility function `getDisplayTransactionType() `in **transactions.ts**:
    * Accepts either an SDK Transaction or Uint8Array
    * For freeze transactions, extracts and returns the specific freeze type string
    * Falls back to standard transaction type for non-freeze transactions
3. Updated components:
    * **TransactionNodeRow.vue** (Ready to Sign, In Progress, Ready for Execution, History tabs in organization mode): Added a watcher that detects freeze transactions and fetches the freeze type on-demand
    * **History.vue** (personal mode): Uses the existing body field from local transactions to extract freeze type
    * **Drafts.vue**: Parses the transactionBytes field to display specific freeze types for draft transactions
    
 **Before**
 
Transaction ID

**After**
Transaction ID


front-end/src/renderer/pages/Transactions/components/History.vue
```diff
@@ -241,6 +242,22 @@ function setPreviousTransactionsIds(id: string | number) {
   }
 }
 
+/**
+ * Gets the display transaction type for local transactions.
+ * For freeze transactions, extracts the specific freeze type from the transaction body.
+ */
+function getLocalTransactionDisplayType(transaction: Transaction): string {
```
jbair06: `getDisplayTransactionType()` does very similar checks and such. Can that function be used in place of `getLocalTransactionDisaplyType()`? Or is this more efficient?

front-end/src/renderer/utils/sdk/transactions.ts
```diff
@@ -122,3 +124,79 @@ export const getFreezeTypeString = (freezeType: FreezeType) => {
...
+ * Gets the display transaction type, including specific freeze types.
+ * For freeze transactions, returns the specific freeze type (e.g., "Freeze Upgrade").
+ * For other transactions, returns the standard type.
+ *
+ * @param transaction - SDK Transaction instance or Uint8Array
+ * @param short - Whether to use short format (no spaces)
+ * @param removeTransaction - Whether to remove " Transaction" suffix
+ * @returns Display string for transaction type
+ */
+export const getDisplayTransactionType = (
```
jbair06: If we could add a type param into this function, then handle both local types and backend types, then could we consolidate all the code a bit more? This should allow us to remove a number of lines of code everywhere else.

jbair06 (changes requested):

front-end/src/renderer/utils/sdk/transactions.ts
```diff
@@ -122,3 +124,79 @@ export const getFreezeTypeString = (freezeType: FreezeType) => {
...
+      return formatTransactionType(freezeTypeStr, short, removeTransaction);
+    }
+  }
+
+  // Fall back to standard transaction type
+  return getTransactionType(sdkTransaction, short, removeTransaction);
+};
+
+// Cache to avoid re-fetching freeze types
+const freezeTypeCache = new Map<number, number | null>();
```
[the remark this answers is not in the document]
ddeskov-limechain: addresed in commits

jbair06 (changes requested):

jbair06 (approved):
Looks good

Issue #2064, opened 2025-12-08
transactions "ready to execute" show "unauthorized" where previously I could see them


ElijahLynn:
### Description

I had the app loaded from Friday, where we saw the transactions, which worked fine, then I tabbed to the app and saw the transactions there, and when I went to details it said unauthorized, then I clicked on all the other tabs and got unauthorized on all of them.

I shutdown the app and re-opened, and all transactions showed again. 

<img width="3044" height="1890" alt="Image" src="https://github.com/user-attachments/assets/fd3af987-711d-4935-a3ac-df9223c8df0a" />

<img width="406" height="274" alt="Image" src="https://github.com/user-attachments/assets/19fb5ead-610b-44c8-ab9f-047f2e1f2420" />

### Steps to reproduce

1. Don't know how to reproduce

### Hedera network

testnet

### Version

v0.18.1

### Operating system

macOS

PR #2185, opened 2026-01-20
"Untauthorized" fixed on tabbing back

ddeskov-limechain:
Fixes issue #2064 

**Problem**
When users tab away from the app and return later, all API requests fail with "Unauthorized" toast errors. The JWT token expires while the app is idle, but there's no mechanism to detect this and trigger re-authentication when the app regains focus.

What Happens:
1. When you're logged in, your JWT token is stored in sessionStorage
2. The backend validates the token on every API request via the `JwtBlacklistGuard` _back-end/apps/api/src/guards/jwt-blacklist.guard.ts_
3. If the token is expired or blacklisted, the backend returns a **401** with `{ message: 'Unauthorized', statusCode: 401 }`
4. The frontend displays this error via toasts (with duration: 0 meaning they persist until dismissed)
Why It Happens After Tabbing Away:
The app has no window focus or visibility change handlers. I searched the entire frontend codebase:
* No visibilitychange event listener
* No window.focus event handling
* No logic to check/refresh tokens when the app regains focus
The only auto-login logic is in _AutoLoginInOrganization.vue_ but it only runs once when organizations are initially loaded - not when the app regains focus.

**Solution**
Add a visibility change listener that validates tokens when the app becomes visible and triggers re-authentication if needed.

Implementation:
1. Create New Composable: _useAppVisibility.ts_ (front-end/src/renderer/composables/useAppVisibility.ts)
Creates a composable that:
* Listens to `document.visibilitychange` events
* Debounces rapid visibility changes (2 seconds)
* Checks if any organization tokens need refresh via shouldSignInOrganization()
* Calls a callback when tokens are expired
2. Modify _AutoLoginInOrganization.vue_ (front-end/src/renderer/components/Organization/AutoLoginInOrganization.vue)
Changes:
* Add `defineExpose()` to expose a `triggerReauthentication()` method
* This method resets the checked flag and calls `openPasswordModalIfRequired()`
3. Integrate in _GlobalAppProcesses.vue_ (front-end/src/renderer/components/GlobalAppProcesses/GlobalAppProcesses.vue)
Changes:
* Import useAppVisibility composable
* Add a template ref to AutoLoginInOrganization
* Use the composable with a callback that triggers re-authentication


svienot (approved):
LGTM

PR #2243, opened 2026-02-09
fix: title of TransactionGroupDetails with only 1 transaction.

svienot:
**Description**:

- Trivial fix for the TransactionGroupDetails page title when the group contains 1 transaction
- (Also 1 isolated reformatting)

**Notes for reviewer**:

Without fix:
Screenshot 2026-02-09 at 11 02 02

With fix:
Screenshot 2026-02-09 at 11 02 25


ericleponner (approved):
LGTM

PR #2322, opened 2026-02-18
feat: test coverage document

### LINKED ISSUE #2282: Identify transaction tool test cases
### Problem

As a tester, I'd like to know what are all the possible flows in the UI are so I can ensure adequate test coverage.

### Solution

* The scope is isolated to the front-end only encompassing user-facing scenarios.
* The PR should contain a single markdown document in the docs folder with a list of test scenarios.
* The document should be broken into high level areas of the application: Login, Upgrade, Settings, Transactions List, Accounts List, Files List, Contact List, etc.
* Within each group it should be further broken down into different areas as appropriate
* Each level should be numbered starting from 1 and sub-levels like 1.1, 1.2.3, etc. so that each test scenario has a distinct number.
* Positive and negative test scenarios should be enumerated for each area.
* Each test scenario should indicate whether this is covered by automated UI testing and whether it has performance tests, if applicable.

### Alternatives

_No response_

Yotsev:
**Description**:

This PR adds a document for the existing automation tests and scenarios that are missing for feature improvement.

**Related issue(s)**:

Fixes #2282


docs/UI_TEST_COVERAGE_AND_RELEASE_CHECKLIST.md
```diff
@@ -0,0 +1,939 @@
...
+
+## 1. Registration
+
+### 1.1 First-Time User Registration
+
+| Check | #     | Scenario                                                         | Automated | Covered By                                               |
+| ----- | ----- | ---------------------------------------------------------------- | --------- | -------------------------------------------------------- |
+|       | 1.1.1 | All registration page elements are displayed correctly           | Yes       | Verify all elements are present on the registration page |
+|       | 1.1.2 | User can register with valid email and strong password           | Yes       | Verify successful registration through "Create New" flow |
+| [ ]   | 1.1.3 | Password strength tooltip shows real-time feedback (>= 10 chars) | No        |                                                          |
```
steven-sheehy: The `[ ]` inside a table doesn't seem to render. Probably we have to remove this column if it's not functional. We can get the same info on what to test based upon the automated=No column.

steven-sheehy (changes requested):
This is a really great start overall. I think it covers most of the positive test cases but it's a missing some negative test cases. An example would be "Attempting to save a transaction group with no name displays an error message".

PR #2335, opened 2026-02-20
fix: Replace the preinstall script with a mustClose directive in electron-builder.yml

ddeskov-limechain:
This PR follows up [PR-2288](https://github.com/hashgraph/hedera-transaction-tool/pull/2288) with an optimization.

**Problem**
The solution for the handling the check if the window was closed when installing a new version was an overkill and the preinstall script was not needed, because it could be simply replaced with a single line of code in `electron-builder.yml`:
```
  pkg:
    mustClose:
      - "com.hiero.transactiontool.app"
```

**Solution**
Tested the approach and it worked. When I have the Transaction Tool App opened and I want to install a new versioin, I got this message:
tallati

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -49,6 +49,10 @@ The Initial Developer of @babel/helper-create-class-features-plugin@7.25.9,
 is The Babel Team (https://github.com/babel/babel).
 Copyright The Babel Team. All Rights Reserved.
 
+The Initial Developer of @babel/helper-globals@7.28.0,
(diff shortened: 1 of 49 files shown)


steven-sheehy (commented):
Need to fix PR to delete preinstall script

ddeskov-limechain:
I have it deleted again. Maybe when we update the branch before merging, it brings the `preinstall` script back, because it is already in `main`.

image

image

steven-sheehy (approved):

PR #2339, opened 2026-02-20
test: add framework for UI unit tests

### LINKED ISSUE #2300: Add framework for UI unit tests
### Problem

We have automation tests that spin up a headless UI to test, but they're heavyweight and tests aren't always added at the same time as the PR that adds the feature. We should have smaller, focused unit tests included in every PR.

### Solution

* Use an existing testing tool/framework
* Use explorer as inspiration as it does this already with vue
* Test individual vue components to ensure given a certain input a certain output is rendered
* Use mocks or stubs for backend calls.
* Add one or two test examples that showcase this approach.
* Future sprints and PRs will slowly increase the unit test coverage.

### Alternatives

_No response_

svienot:
**Description**:

The vitest framework is already used in the project for the existing front-end unit tests. Among those:
- the tests of the **main** components are the only ones run in the CI.
- several tests in the **renderer** component are covering utilities but these are never executed in CI (they aren’t included in any test).
- there is one existing test file for the **shared** component of the frontend which is never executed in CI either (and it is failing)

In this PR we:
- leverage the vitest framework and extend its usage to UI components of the renderer component:
  - add `@pinia/testing`, `@vue/test-utils` and `happy-dom` dependencies
  - add vitest configuration for the **shared** component, which was not yet present
  - skip the failing tests in the **shared** component, for the time being
  - update the README with the new `test:renderer` and `test:shared` npm targets
  - update the `test-frontend.yaml` CI workflow to include the renderer and shared components unit tests in the test matrix. We do not however include those in the coverage measurement.
- and also add a unit test for a simple UI component (BreadCrumb) as a first example.

**Related issue(s)**:

Fixes #2300 

**Checklist**


ericleponner (approved):
LGTM

steven-sheehy (commented):
It looks good for view only tests, but shouldn't we also have an example that calls a remote API like MN, CN, or TT backend? Should these be follow ups? Or is that logic already isolated in non-vue components?

.github/workflows/test-frontend.yaml
```diff
@@ -210,7 +214,7 @@ jobs:
         working-directory: front-end
 
       - name: Run Unit Tests
-        run: pnpm run test:${{ matrix.test-suite.command }}:coverage
+        run: pnpm run test:${{ matrix.test-suite.command }}
```
steven-sheehy: Why do we not include coverage? We want all areas to be calculated so we know what's missing.

svienot:
> It looks good for view only tests, but shouldn't we also have an example that calls a remote API like MN, CN, or TT backend? Should these be follow ups? Or is that logic already isolated in non-vue components?

Yes, next step should be to add a more substantial test, using a mock-up of an API call (to MN and/or the backend). 
Perhaps we can integrate this as is for the time being and follow up.

steven-sheehy (approved):

PR #2360, opened 2026-02-24
feat: set 'sign next' as the default action in signing process

### LINKED ISSUE #2359: User feedback: make "Sign & Next" the default signing action
**Description**  
Users have requested that the primary signing button default to **"Sign & Next"** instead of just **"Sign"**. The current behavior requires extra clicks the first time a user is signing multiple transactions in a row.

**Current behavior**  
- The main action button on the signing screen is labeled **"Sign"**.  
- Users must take an additional action (e.g., use a secondary button or navigation) to move to the next transaction after signing.

**Requested behavior**  
- Make **"Sign & Next"** the default/primary action button when signing transactions.

jbair06:
**Description**:
This pull request makes small but important adjustments to the transaction signing flow in the front-end. The main changes ensure that after signing a transaction, the navigation logic correctly handles whether to proceed to the next transaction or return to the transaction list, and updates the default behavior for navigation after signing.

Transaction navigation improvements:

* Changed the default value of `goToNext` in `SplitSignButtonDropdown.vue` to `true`, so users will now automatically proceed to the next transaction after signing by default.
* Updated the `handleSign` function in `TransactionDetailsHeader.vue` to check if there is a next transaction available; if not, it routes the user back to the transaction list instead of attempting to navigate to a non-existent transaction.

**Related issue(s)**:

Fixes #2359 

**Notes for reviewer**:
When 'Sign & Next' on a Transaction inside a Transaction Group, note that it routes back to the Transaction Group details.

**Checklist**

### DIFF
--- a/.github/workflows/pr-formatting.yaml
+++ b/.github/workflows/pr-formatting.yaml
@@ -28,7 +28,7 @@ jobs:
     runs-on: transaction-tools-linux-medium
     steps:
       - name: Harden Runner
-        uses: step-security/harden-runner@5ef0c079ce82195b2a36a210272d6b661572d83e # v2.14.2
+        uses: step-security/harden-runner@a90bcbc6539c36a85cdfeb73f7e2f433735f215b # v2.15.0
         with:
           egress-policy: audit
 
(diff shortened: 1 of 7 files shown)


ddeskov-limechain:
LGTM

PR #2338, opened 2026-02-20
fix: automatically set notifications read after signing

jbair06:
**Description**:
This pull request introduces a new bulk update mechanism for notification receivers when transactions are signed, improves backend transaction processing, and simplifies related frontend notification handling. The backend now marks relevant notifications as read in bulk, and the frontend removes redundant notification logic, relying on backend updates.

**Backend enhancements:**

* Added `bulkUpdateNotificationReceivers` method to `SignersService`, enabling efficient bulk marking of notifications as read when transactions are updated.
* Integrated notification updates into the transaction signing flow: notifications are collected and updated alongside transaction bytes, ensuring consistency. [[1]](diffhunk://#diff-45a2f3f118a08ffd6b7d3f4cecd2d83e838d56cde6cebdec2b57c3ff51d23d75R277) [[2]](diffhunk://#diff-45a2f3f118a08ffd6b7d3f4cecd2d83e838d56cde6cebdec2b57c3ff51d23d75R296) [[3]](diffhunk://#diff-45a2f3f118a08ffd6b7d3f4cecd2d83e838d56cde6cebdec2b57c3ff51d23d75R320-R324)
* Comprehensive tests for `bulkUpdateNotificationReceivers`, verifying query execution, parameter handling, and conditional behavior.
* Improved test naming for clarity in bulk update scenarios.

**Frontend simplification:**

* Removed notification store imports and notification marking logic from `TransactionNodeRow.vue`, as notifications are now handled by the backend. [[1]](diffhunk://#diff-bcd22789c7c11313f2fab68765b804f11f579ade8c0f653c405488084a059e7bL8) [[2]](diffhunk://#diff-bcd22789c7c11313f2fab68765b804f11f579ade8c0f653c405488084a059e7bL44) [[3]](diffhunk://#diff-bcd22789c7c11313f2fab68765b804f11f579ade8c0f653c405488084a059e7bL189-L209)
* Updated event handlers to directly emit transaction/group signed events, further streamlining the component.

**Related issue(s)**:

Fixes #2328

**Checklist**

- [x] Tested (unit, integration, etc.)

PR #2367, opened 2026-02-24
feat: auto read messages after signing

jbair06:
**Description**:

When a user signs the transaction, the notifications for that message are now auto set to 'isRead=true' if applicable. 

Cherry pick of #2338


jbair06:
Wrong branch.

PR #2504, opened 2026-03-18
fix: handle explicit time zones in automation

yasenltd:
**Description**:
- only append `Z` when the timestamp has no explicit timezone
- skip empty valid-start values and improve wait logging/error context
- refresh `NOTICE` for updated dependency metadata

This PR modifies the automation valid-start wait helper in order to support both Hedera UI timestamps and ISO timestamps without incorrectly rewriting values that already include a timezone. It also makes the helper more defensive for empty inputs and updates generated third-party dependency metadata in `NOTICE`.

**Related issue(s)**:

Fixes #

**Notes for reviewer**:
- Main functional change is in `automation/utils/util.ts`
- `waitForValidStart` now preserves explicit timezone suffixes such as `Z`, `GMT`, `+HH:MM`, and `+HHMM`
- No automated tests were run as part of this change

**Checklist**


ericleponner (approved):
LGTM

Issue #2518, opened 2026-03-19
Prevent duplicate emails/notifications from multiple backend pods


jbair06:
**Description**:  
Some emails and notifications are being sent multiple times. When a transaction expires or gathers enough signatures to become ready for execution, duplicate emails are received, and this has been observed on every backend where more than one `api/chain` pod is running.

**Requirements**:  
- Ensure that in multi-pod deployments, each notification is sent **exactly once** per event, even when multiple `api/chain` pods are running.

PR #2534, opened 2026-03-25
fix: multiple backend pods notifications

Kolpic:
Fixes [#2518](https://github.com/hashgraph/hedera-transaction-tool/issues/2518)

**Description**:
The problem is that each pod runs the same cron jobs on the same schedule. When a transaction expires, all pods independently detect it and each publishes a notification event to NATS. The notifications service processes all of them, resulting in duplicate emails — one per pod.
NATS delivers messages exactly as it's told to. The problem is 3 pods all telling NATS to deliver the same message. By the time NATS is involved, it's too late — the duplicates are already created so the fix needs to happen before the NATS publish — at the cron job level.

I reproduced this locally by running `docker compose up --scale chain=3`. The logs showed:

> chain-1  | Published to: notifications.queue.transaction.status-update, seq: 2
chain-2  | Published to: notifications.queue.transaction.status-update, seq: 3
chain-3  | Published to: notifications.queue.transaction.status-update, seq: 4
notifications-1  | Processing email batch... with 3 notifications in 1 groups

**The Solution**
I used Redis as a shared coordinator. Before each cron job runs its logic, it tries to set a key in Redis with the `NX` flag (only set if not exists). Only the first pod to reach Redis succeeds — the others see the key already exists and skip.

**Result After the Fix**

> chain-1  | [CronLock] Acquired lock for status_update_expired_transactions
chain-2  | [CronLock] Skipping status_update_expired_transactions — another pod holds the lock
chain-3  | [CronLock] Skipping status_update_expired_transactions — another pod holds the lock
chain-1  | Published to: notifications.queue.transaction.status-update, seq: 50
notifications-1  | Processing email batch... with 1 notifications in 1 groups
notifications-1  | Message sent: <6d753e6e-...@limechain.tech>


back-end/apps/chain/src/transaction-scheduler/transaction-scheduler.service.ts
```diff
@@ -20,23 +22,52 @@ import {
...
+  /**
+   * Tries to acquire a distributed lock using Redis SET NX.
+   * Returns true if the lock was acquired, false if another pod already holds it.
+   */
+  private async tryAcquireCronLock(key: string, ttlMs: number): Promise<boolean> {
+    try {
+      const result = await this.redis.set(`cron:${key}`, '1', 'PX', ttlMs, 'NX');
+      return result === 'OK';
+    } catch (error) {
+      console.error(`[CronLock] Error acquiring lock for ${key}:`, error.message);
```
steven-sheehy: Use the logger

PR #2672, opened 2026-04-14
fix: preserve transaction description for multiple account update groups

### LINKED ISSUE #2546: Preserve entered transaction description for multiple account update groups
**Description**:  
In the account update transaction creation flow, users can update multiple accounts at once by selecting **Multiple Accounts** and entering a range of accounts. This creates a transaction group of account update transactions, but the group currently gets an auto-generated description. If the user enters a value in the **Transaction Description** field during creation, that value is lost. The transaction group should use the entered transaction description as the group description.

**Requirements**:  
- When **Multiple Accounts** is selected and a transaction group is created, capture the value entered in the **Transaction Description** field.  
- Apply that value to the resulting transaction group as its description.  
- Do not overwrite the user-entered description with an auto-generated one for this flow.  
- Keep existing behavior unchanged for single-account account update transactions.

**Acceptance criteria**:  
- A user-entered transaction description is preserved when creating a multiple-account update group.  
- The resulting transaction group displays the entered description instead of an auto-generated description.  
- Single-account flows continue to behave as they do today.

svienot:
**Description**:

When the user checks _Multiple accounts_ in the account update transaction creation flow, a group of Account Updates is automatically created.  With the changes below:
- if the user enters a value in the Transaction Description field, this value is used as the group description (instead of
- if the user  does not provide a description, we keep the default value previously used in all cases

**Related issue(s)**:

Fixes #2546

### DIFF
--- a/front-end/src/renderer/components/Transaction/Create/BaseTransaction/BaseTransaction.vue
+++ b/front-end/src/renderer/components/Transaction/Create/BaseTransaction/BaseTransaction.vue
@@ -192,6 +192,7 @@ const handleCreate = async () => {
     processable.payerId = payerData.accountId.value;
     processable.baseValidStart = data.validStart;
     processable.maxTransactionFee = data.maxTransactionFee as Hbar;
+    processable.description = description.value;
   }
 
   await withLoader(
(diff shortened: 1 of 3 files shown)


ericleponner (approved):
LGTM

PR #2674, opened 2026-04-14
chore: add build:mac:arm64 target

svienot:
**Description**:

One line change to `package.json` to add the build target `build:mac:arm64` -- to build only for arm64 architecture in development since the full electron-build is painfully slow.


ericleponner (approved):
LGTM

PR #2675, opened 2026-04-14
fix: app menu highlighting

svienot:
**Description**:

One-line fix to make sure router.appMenuItem is updated when user selects Settings.

Fixes #2612


ericleponner (approved):
LGTM

Issue #2521, opened 2026-03-20
Preload next transaction data for faster “Sign & Next” navigation


jbair06:
**Description**:  
Users report that loading the next transaction after using “Sign & Next” or navigation controls takes longer than expected. The transition to the next transaction should feel nearly instant and app-like. To achieve this, we need to preload not only the transaction node data but all required transaction details for the next item before the user navigates, or a similar solution.

**Requirements**:  
- Investigate current data loading flow for:
  - “Sign & Next” actions.  
  - Next/previous navigation between transactions.  
- Implement preloading of the next (and possibly previous) transaction’s full data set, not just transaction node metadata, while the user is viewing the current transaction.  
- Ensure the UI shows the next transaction immediately on “Sign & Next” or navigation, with minimal or no visible loading state.  
- Avoid redundant network calls by reusing preloaded data when possible.  

**Acceptance criteria**:  
- For typical transaction sizes, “Sign & Next” and next/previous navigation feel near-instant to users, with no long-loading spinner between transactions.    
- Preloading does not introduce correctness issues (e.g., stale data) and falls back gracefully if prefetch fails.

PR #2676, opened 2026-04-14
fix: caching  GET /transactions/{transactionId} (addendum)

ericleponner:
**Description**:

This PR is a follow-up of #2626.

Changes below add some logic for clearing transaction cache when notifications are received.
When a `TRANSACTION_NOTIFICATION` arrives, `EntityCache.forget()` is called to remove existing data from BackendTransactionCache. However this removal is conditional : data are removed only if they are 1s or more old.
This behavior enables to avoid clearing the cache when notification is received immediately after a full refresh.

**Related issue(s)**:

Contributions to #2521

**Notes for reviewer**:

```
M   front-end/src/renderer/caches/base/EntityCache.ts
    # Added 'strict' parameter to EntityCache.forget()
    # When strict == true, forget() behavior is unchanged : entry is removed from cache.
    # When strict == false, forget() removes data if it is older than 1s
    # Default value is strict = true.

M   front-end/src/renderer/pages/TransactionDetails/TransactionDetails.vue
M   front-end/src/renderer/pages/TransactionGroupDetails/TransactionGroupDetails.vue
M   front-end/src/renderer/pages/Transactions/components/History.vue
M   front-end/src/renderer/pages/Transactions/components/TransactionNodeTable.vue
    # In useWebsocketSubscription(), those view now invoke a specialized fetch():
    # fetchXXXOnNotif() calls regular fetchXXX() but after calling BackendTransactionCache.forget().
```

**Checklist**

### DIFF
--- a/front-end/src/renderer/caches/base/EntityCache.ts
+++ b/front-end/src/renderer/caches/base/EntityCache.ts
@@ -28,9 +28,15 @@ export abstract class EntityCache<K extends string | number, E> {
     return result;
   }
 
-  public forget(key: K, mirrorNodeUrl: string): void {
+  public forget(key: K, mirrorNodeUrl: string, strict = true): void {
     const recordKey = this.makeRecordKey(key, mirrorNodeUrl);
-    this.records.delete(recordKey);
+    const currentRecord = this.records.get(recordKey);
+    if (currentRecord) {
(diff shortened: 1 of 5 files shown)


svienot (approved):
LGTM

PR #2723, opened 2026-04-21
fix: hide file hash row when hash is empty

### LINKED ISSUE #2586: FreezeAbort transaction incorrectly displays `filehash` as `0x`
**Description:**  
When viewing **FreezeAbort** transaction details, the UI currently shows a `file hash` field with the value `0x`. This field should **not** be displayed, as `FreezeAbort` transactions do not include or use a file hash.

**Expected Behavior:**  
- The `file hash` field should be removed for `FreezeAbort` transactions.

**Actual Behavior:**  
- A `file hash` field appears with the placeholder value `0x`.

Kolpic:
Fixes: [#2586](https://github.com/hashgraph/hedera-transaction-tool/issues/2586)

**Description**:
When viewing FreezeAbort transaction details, the UI currently shows a file hash field with the value 0x. This field should not be displayed, as FreezeAbort transactions do not include or use a file hash.

**Cause**
In `front-end/src/renderer/components/Transaction/Details/FreezeDetails.vue`, the template guards the File Hash row with:

> <div v-if="transaction.fileHash" class="col-12 py-3">

The Hedera SDK's `FreezeTransaction.fileHash` getter returns a Uint8Array — never null or undefined. For freeze types that don't set a hash (including FreezeAbort), it returns an empty Uint8Array(0), because that's the protobuf default for a bytes field.

In JavaScript, an empty Uint8Array is truthy (it's an object, not null), so:

- `v-if="transaction.fileHash" `passes → the row renders.
- onBeforeMount runs uint8ToHex(transaction.fileHash), which iterates zero bytes and returns `""`.
- The template renders `0x${fileHashHex}` → `0x${""}` → `0x`.
The transaction creation logic in `createTransactions.ts` correctly skips setFileHash for FreezeAbort, so the bug is purely on the display side.

**Fix**
Change the guard to also check the array's length, so an empty Uint8Array is treated as "no hash set":

> <div v-if="transaction.fileHash && transaction.fileHash.length > 0" class="col-12 py-3">

This hides the File Hash row for FreezeAbort (and any other freeze type that doesn't populate a hash), while still rendering it correctly for PrepareUpgrade and FreezeUpgrade where a real hash is present.

### DIFF
--- a/front-end/src/renderer/components/Transaction/Details/FreezeDetails.vue
+++ b/front-end/src/renderer/components/Transaction/Details/FreezeDetails.vue
@@ -59,7 +59,7 @@ const commonColClass = 'col-6 col-lg-5 col-xl-4 col-xxl-3 overflow-hidden py-3';
     </div>
 
     <!-- File Hash -->
-    <div v-if="transaction.fileHash" class="col-12 py-3">
+    <div v-if="transaction.fileHash && transaction.fileHash.length > 0" class="col-12 py-3">
       <h4 :class="detailItemLabelClass">File Hash</h4>
       <p :class="detailItemValueClass">
         {{ fileHashHex.startsWith('0x') ? fileHashHex : `0x${fileHashHex}` }}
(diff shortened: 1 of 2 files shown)


steven-sheehy:
Please add a unit test

ericleponner (approved):
LGTM

Issue #2552, opened 2026-03-28
Improve signing flow and error handling


jbair06:
**Description:**  
Update the signing flow so **Sign & Next** only advances after a successful signing operation. Replace the generic **failed to connect** error with a more actionable error message and add logging that provides enough detail to diagnose signing failures.

**Requirements:**  
- Ensure **Sign & Next** only proceeds after signing succeeds.  
- Replace the generic **failed to connect** message with a clearer, more actionable error.  
- Add logging that helps identify why signing failed.  
- Add tests or verification for the failure path.

**Acceptance criteria:**  
- The UI does not advance unless signing completes successfully.  
- Users see a useful error when signing fails.  
- Logging provides enough detail to troubleshoot the failure.  
- The signing flow is covered by tests or reproducible validation.

jbair06:
In a scenario like this, the user should also be presented with instructions in addition to the error. Something like, 'connection is lost, try again' or similar.

ericleponner:
With #2607, #2609 and #2631 we can close this issue.

PR #2755, opened 2026-04-29
chore(transactions): batch UserKey lookups across DTOs on signature i…

Kolpic:
## Summary

Follow-up to PR #2745 (issue #2552). Replaces the N+1 UserKey query in `importSignatures` with a single bulk lookup. **_No behavioral change._**

## Problem

`importSignatures` issued one `entityManager.find(UserKey, ...)` per DTO. A 50-DTO batch hit the DB 50 times. Flagged by Copilot review on #2745, deferred to keep that fix scoped.

## Change

Restructured into two passes around one bulk query:

1. **Pass 1** — validate each DTO, mutate signatures, collect every public key (raw + DER) into a global Set.
2. **Bulk lookup** — one `find(UserKey, { where: { publicKey: In([...]) } })`, results bucketed into `Map<string, UserKey[]>`.
3. **Pass 2** — assemble signer rows per DTO from the in-memory map.

Buckets are arrays because `UserKey.publicKey` is indexed but not unique — two users can share a key, both rows must surface.

## Tests

- Single bulk lookup across multiple DTOs.
- Shared-publicKey rows both produce signer rows.
- Soft-deleted UserKey is excluded (`withDeleted` never set).

All existing tests still pass.

## Manual verification

User A creates a multi-sig tx, exports TTv1, then user B signs, User A imports the zip, User B logs in and sees full access.


jbair06 (approved):
LGTM

PR #2779, opened 2026-05-04
feat: use pnpm workspaces

### LINKED ISSUE #2751: Use PNPM workspaces
### Problem

Currently we get separate dependabots for each module even if the same dependency is used in multiple modules.

### Solution

* Add a `pnpm-workspace.yaml` in root folder to tell which pnpm which dirs are packages
* Update root `package.json` to hoist shared dependencies from sub-dirs
* Remove shared dependencies from child modules
* Update CI workflows, docs, etc to run pnpm install at root

### Alternatives

_No response_

Kolpic:
Fixes: [2751](https://github.com/hashgraph/hedera-transaction-tool/issues/2751)

# Summary

The repo had three independent pnpm workspaces: `back-end/`, `front-end/`, `automation/`. Each shipped its own `package.json` and `pnpm-lock.yaml`.

This caused four real problems:

- **Dependabot triplication** — every shared-dep bump opened 3 PRs.
- **Version drift** — shared deps could diverge silently between modules.
- **Slow CI** — three `pnpm install` calls per pipeline.
- **Cognitive overhead** — contributors had to know which directory to install in.

Goal: one root workspace, one lockfile, one catalog of shared versions.

# What changed

## Root workspace + catalog

- New `pnpm-workspace.yaml` at the repo root declares all packages.
- New root `package.json` holds a single dependency catalog.
- Child `package.json` files reference the catalog:
  ```json
  "axios": "catalog:"

- Deleted three child lockfiles and the old `back-end/pnpm-workspace.yaml.`
- Result: pnpm install at the root installs everything. `pnpm -F <name> <script>` runs scripts in any child.

## Dependabot consolidation
- `.github/dependabot.yml:` 4 ecosystems → 1 npm ecosystem at the root + github-actions.
- Added groups (hiero-ledger, nestjs, testing) so related bumps land in a single PR.
- This is the direct fix for the issue.

## CI workflows
Updated 4 workflows (test-backend, test-frontend, test-migrations, publish-docker-images):

- Install at the repo root, not per-module.
- Cache against the root `pnpm-lock.yaml`.
- Run scripts via `pnpm -F <module> <script>.`
- Docker build context flipped to `.` (repo root).

## Dockerfiles
Rewrote 4 Dockerfiles (api, chain, notifications, typeorm) on a consistent recipe:

- Build context is the repo root.
- Copy all workspace manifests + the root lockfile before install (keeps the install layer cacheable).
- Filtered install: `pnpm install --filter "back-end..." --filter "@back-end/api..."` — only what each service needs.
- Multi-stage: development builds; production uses `--prod --frozen-lockfile` and copies dist from development.

## Compose + deploy automation
- Three compose files updated: context: `..` so they resolve to the repo root.
- Dev volume mount adjusted to `:/usr/src/app/back-end` so hot-reload still works.
- `back-end/k8s/dev/shell/docker_images.sh:` added one extra `..` to land at the repo root.
- `cloudbuild.yaml:` prefixed every Dockerfile path with `back-end/`.
- Added a root `.dockerignore` so the new (larger) build context stays lean.

## Verification
12-step manual checklist, all green:

Num | Check | Result
-- | -- | --
1 | Single root workspace | ✓
2 | pnpm install --frozen-lockfile at root | ✓
3 | Catalog refs resolve | ✓
4 | All three modules build | ✓
5 | Tests pass | ✓ 
6 | pnpm -F works from any cwd | ✓
7 | Direct Docker builds (4 images) | ✓
8 | docker-compose dev mode | ✓
9 | docker-compose prod mode | ✓
10 | Deploy script paths | ✓
11 | Dependabot 4 → 1 ecosystems | ✓

## Final layout

```
hedera-transaction-tool/
├── package.json              # root manifest (catalog)
├── pnpm-workspace.yaml       # workspace declaration
├── pnpm-lock.yaml            # THE lockfile
├── .dockerignore
├── back-end/
│   ├── package.json          # name: back-end, catalog: refs
│   ├── apps/{api,chain,notifications}/Dockerfile  # context = repo root
│   ├── typeorm/Dockerfile
│   └── docker-compose*.yaml  # context: ..
├── front-end/
│   ├── package.json          # name: front-end (renamed)
│   └── electron-builder.yml  # hardcoded artifact prefix
└── automation/
    └── package.json          # name: automation (renamed)

```

    
One workspace. One lockfile. One catalog. One Dependabot npm ecosystem. One install command.


back-end/typeorm/Dockerfile
```diff
@@ -36,17 +40,25 @@ RUN npm install -g pnpm
...
+COPY --from=builder /usr/src/app/back-end/apps/api/package.json ./back-end/apps/api/package.json
+COPY --from=builder /usr/src/app/back-end/apps/chain/package.json ./back-end/apps/chain/package.json
+COPY --from=builder /usr/src/app/back-end/apps/notifications/package.json ./back-end/apps/notifications/package.json
 
 ENV CI=true
+ENV SKIP_NOTICE_GEN=true
 
-# Install only typeorm production dependencies
-RUN pnpm install --prod --filter typeorm --ignore-scripts
+RUN pnpm install --prod --filter typeorm --frozen-lockfile --ignore-scripts
```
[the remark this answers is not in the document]
Kolpic: Correct that this can be fragile. The entities under `libs/common/ `currently only import from `typeorm`, so the existing `--filter typeorm` works today, but any future class-validator or `@nestjs/swagger` decorator on an entity would break the migration container. Will aligning with how `api/chain/notifications` do it: `--filter "back-end..." --filter "typeorm...".`

.github/workflows/test-frontend.yaml
```diff
@@ -42,14 +42,13 @@ jobs:
           node-version: 22.12.0
           cache: 'npm'
```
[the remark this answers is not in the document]
Kolpic: Correct — but i think that the one-line swap to cache: 'pnpm' won't work on its own though, because setup-node would call pnpm store path before pnpm is installed. Switching to `pnpm/action-setup` before `actions/setup-node` so the pnpm store cache actually engages — applying across all four workflow files.

package.json
```diff
@@ -1,5 +1,29 @@
...
-    "prettier": "^3.8.3"
+    "prettier": "catalog:"
+  },
+  "pnpm": {
+    "overrides": {
+      "protobufjs@<8.0.0": "7.5.5",
+      "protobufjs@>=8.0.0": "8.0.1",
+      "follow-redirects": "1.16.0",
+      "lodash": "4.18.0",
+      "handlebars": "4.7.9",
```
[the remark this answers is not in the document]
Kolpic: Correct and double checked — lodash@4.18.0 is marked deprecated on the npm registry with the message "Bad release. Please use lodash@4.17.21 instead." Fixing the override to 4.17.21.

.github/workflows/test-frontend.yaml
```diff
@@ -96,7 +95,7 @@ jobs:
...
```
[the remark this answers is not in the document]
Kolpic: Correct — but i think that the one-line swap to cache: 'pnpm' won't work on its own though, because setup-node would call pnpm store path before pnpm is installed. Switching to `pnpm/action-setup` before `actions/setup-node` so the pnpm store cache actually engages — applying across all four workflow files.

front-end/README.md
```diff
@@ -45,25 +45,35 @@ and is at your own risk.
...
+
 ```bash
 pnpm install
 ```
 
+Front-end-specific scripts can then be run via the workspace filter (from any directory):
+
+```bash
+pnpm -F hedera-transaction-tool <script>          # e.g. pnpm -F hedera-transaction-tool dev
+```
```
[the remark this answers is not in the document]
Kolpic: Correct - i will update the pr description, since there is no renamed packages.

automation/README.md
```diff
@@ -17,25 +17,29 @@ This folder contains automated test tooling for Hedera Transaction Tool:
...
    ```bash
+   cd automation
    cp example.env .env
    ```
 
+Automation-specific scripts can be run via the workspace filter from any directory:
+
+```bash
+pnpm -F hedera-transaction-tool-tests-v2 <script>
+```
```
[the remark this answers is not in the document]
Kolpic: Correct - i will update the pr description, since there is no renamed packages.

.dockerignore
steven-sheehy, line 1: Should delete the back-end dockerignore

package.json
```diff
@@ -1,5 +1,29 @@
...
   "devDependencies": {
-    "prettier": "^3.8.3"
+    "prettier": "catalog:"
+  },
+  "pnpm": {
+    "overrides": {
+      "protobufjs@<8.0.0": "7.5.5",
+      "protobufjs@>=8.0.0": "8.0.1",
+      "follow-redirects": "1.16.0",
+      "lodash": "4.17.21",
```
steven-sheehy, line 12: Some modules were using 4.18. Can we use the latest everywhere so we don't regress any vulnerabilities?

jbair06: According to this, 4.18.0 is a bad release. 

https://www.npmjs.com/package/lodash/v/4.18.0

We could pin 4.18.1 instead, or leave it for another PR?

package.json
jbair06, line 1: Should we add -
```
"engines": {
    "node": ">=22.12.0"
  },
```
here? Perhaps unnecessary, but a bit helpful for onboarding, etc.

front-end/package.json
```diff
@@ -36,29 +36,31 @@
   },
   "dependencies": {
     "@electron-toolkit/utils": "3.0.0",
-    "@hiero-ledger/proto": "2.28.0",
-    "@hiero-ledger/sdk": "2.82.0",
+    "@hiero-ledger/cryptography": "1.18.0",
```
jbair06, line 39: Do we need this? I thought all the cryptography stuff we use is in hiero-ledger/sdk.

Kolpic: yes, it comes in transitively via the sdk the direct declaration is redundant i will remove.

package.json
Kolpic: I will add it

package.json
Kolpic: All are using the `4.17.21` version. I am not sure what is changed in the `4.18.1` i have to check. I think it will be better to leave the `4.18.1` for another pr

front-end/package.json
Kolpic: Actually we do need it — removing the direct dep breaks `vue-tsc -p tsconfig.web.json `even though there are zero runtime imports. Tried removing it in a follow-up commit and CI failed with build error. I think that even though front-end never imports `@hiero-ledger/cryptography` directly at runtime, the sdk re-exports types from it (like the PrivateKey appears in the inferred return types of useAccountId, useNodeId, and several components) and TypeScript's portability check requires that any type used in a public signature come from a package that's _**directly declared**_ in the manifest.

jbair06 (approved):

steven-sheehy (approved):

PR #2797, opened 2026-05-07
fix: app password leave dialog flow in a stale empty state

### LINKED ISSUE #2749: App password dialog can leave change password dialog flow in a stale empty state
**Description:**
When I had an org selected, I went to **Settings -> Profile** and tried to change my password by entering my old and new password. I clicked **Change password**, then canceled the confirmation dialog, and clicked **Change password** again. After accepting the change and entering my app password when prompted, the flow failed with **“password cannot be empty”** even though the password fields appeared non-empty at that point.

I was able to reproduce this even without canceling the confirmation dialog first, so the issue seems tied to the change-password flow after the app-password prompt appears. Electron apps can be sensitive to stale UI state after modal dialogs or authentication prompts, which can leave inputs in an unexpected state if the flow is restarted.

**Steps to reproduce:**
1. Select an org.
2. Go to **Settings -> Profile**.
3. Enter old password and new password.
4. Click **Change password**.
7. Accept the change.
8. Enter the app password when prompted.

**Actual result:**
The app errors with **“password cannot be empty”**, and the password fields are empty.

**Expected result:**
The password change should proceed normally after the app password prompt is completed.

**Notes:**
- I could reproduce it even without first canceling the confirmation dialog.
- The issue happens after a sufficient time has passed that the app requests the app password, when available when not using keychain.

Kolpic:
Fixes: [2749](https://github.com/hashgraph/hedera-transaction-tool/issues/2749)

## Summary

The change-password flow in **Settings → Profile** could be left in a broken state after a failed attempt — fields appeared empty, retries returned `"Password cannot be empty"`, and on a keychain Deny the backend was rotated to the new password while the local credentials store kept the old one.

## Root causes

1. `handleChangePassword` cleared the password fields in its `finally` block — including the early-return path when the personal-password modal opened. The modal's submit callback then re-invoked the handler with empty fields.
2. The backend password rotation ran **before** the local keychain encryption. Denying the keychain prompt failed only the local step, leaving backend and local DB out of sync.

## Fix

- Moved the field reset onto the success path; `finally` only resets the loading flag.
- New `encryptOrganizationPassword` IPC + service runs the keychain prompt in isolation. The handler now encrypts first → rotates the backend → writes the already-encrypted blob locally via a new `passwordIsEncrypted` flag on `updateOrganizationCredentials`. A keychain Deny aborts before the backend is touched.
- Three distinct failure-mode messages (`"Keychain access denied or unavailable"`, `"Failed to encrypt with application password"`, `"No encryption method available"`) reach the toast with the original error preserved as `cause`.

## Tests

Unit tests across `organizationCredentials.spec.ts` and `SettingsProfile.spec.ts` covering the encrypt-first ordering, all three failure modes, the form-state preservation, the personal-password modal callback re-invocation, and the no-method-available `logger.warn` for observability.

## Out of scope — tracked separately

After a macOS keychain Deny, all subsequent `safeStorage.encrypt`/`decrypt` calls fail until the app restarts (a Chromium + macOS Keychain Services behavior). This affects org login, auto-sign-in, and key signing — a follow-up sub-issue will address the recovery UX.

### DIFF
--- a/front-end/src/main/modules/ipcHandlers/localUser/organizationCredentials.ts
+++ b/front-end/src/main/modules/ipcHandlers/localUser/organizationCredentials.ts
@@ -7,6 +7,7 @@ import {
   deleteOrganizationCredentials,
   tryAutoSignIn,
   getOrganizationCredentials,
+  encryptOrganizationPassword,
 } from '@main/services/localUser';
 import { createIPCChannel, renameFunc } from '@main/utils/electronInfra';
 
@@ -21,5 +22,6 @@ export default () => {
     renameFunc(deleteOrganizationCredentials, 'deleteOrganizationCredentials'),
     renameFunc(getOrganizationCredentials, 'getOrganizationCredentials'),
     renameFunc(tryAutoSignIn, 'tryAutoSignIn'),
+    renameFunc(encryptOrganizationPassword, 'encryptOrganizationPassword'),
   ]);
 };
--- a/front-end/src/main/services/localUser/organizationCredentials.ts
+++ b/front-end/src/main/services/localUser/organizationCredentials.ts
@@ -291,12 +291,8 @@ export const tryAutoSignIn = async (user_id: string, decryptPassword: string | n
   return failedLogins;
 };
 
-/* Encrypts an organization password for local storage.
- * Surfaces keychain-deny / wrong-personal-password failures up front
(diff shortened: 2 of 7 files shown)


jbair06 (approved):
Looks good

PR #2799, opened 2026-05-07
fix: stepper state status

### LINKED ISSUE #2796: TransactionDetails stepper renders stale status on mount
### Description

On the transaction detail page, status-changing WebSocket events that fire before TransactionDetails.vue mounts are silently dropped, because the useWebsocketSubscription(TRANSACTION_ACTION, …) subscriber only exists while the component is mounted. The cached backend transaction therefore stays at its pre-event status, and the stepper renders that stale value on first paint.

### Steps to reproduce

1. Be on a list view (e.g. Ready to Sign) when an upstream actor signs/executes a transaction such that a TRANSACTION_ACTION WS event fires while the detail page is not mounted.
2. Open the transaction detail page.
3. Observe: stepper shows the prior status (e.g. "Waiting for signatures") even though the backend now reports the new status (e.g. "Executed").
4. Forcing a refetch (navigating away and back, or invalidating the cache) clears the stale render.

### Additional context

Affected file                                                                                                                                                                
- `front-end/src/renderer/pages/TransactionDetails/TransactionDetails.vue `— onBeforeMount currently calls `fetchTransaction()`, which reads from the cache and skips invalidation.
                                                                                                                                                                               
Proposed fix                                                                                                                                                                 
- Invalidate the cached entry once on mount before the first fetch — the same `fetchTransactionOnNotif()` path the WS handler already uses:                                      
                                                                                                                                                                               
```
  onBeforeMount(async () => {
    const id = formattedId.value;                                                                                                                                              
    if (!id) { router.back(); return; }                                                                                                                                        
    await fetchTransactionOnNotif(); // invalidate cache, then fetch
  });          
```                                                                                                                                                                
                             
This was originally bundled into PR #2783 but reverted at review time so that PR could stay scoped to the automation refactor.                                               
                             
Test impact: Two regression tests in automation/tests/organization-advanced/organizationTransactionLifecycleTests.test.ts depend on this behavior and have been temporarily skipped pending this fix:                                                                                                                                                                         
  - Verify transaction is shown "Ready for Execution" and correct stage is displayed
  - Verify transaction is shown "History" after it is executed                                                                         

Both tests should be re-enabled (drop the .skip) as part of the PR that resolves this issue.                                                                                 
                                                                                                                                                                               
Related
  - PR #2783 (where the original fix was proposed and reverted)

### Hedera network

_No response_

### Version

0.30.0

### Operating system

None

ericleponner:
**Description**:

Changes below fix cache handling.

`App` view now listens to `TRANSACTION_ACTION` notifications and make sure that target transactions are evicted from backend transaction cache. This enables cache to be maintained up to date even when `Transaction Details` / `Transaction Group Details` pages are not displayed.

Even with those changes, there are two automation tests from `organizationTransactionLifecycleTests.test.ts` which refuse to work.
They will be investigated in a separate task (see #2804).

**Related issue(s)**:

Fixes #2796

**Notes for reviewer**:

```
M   front-end/src/renderer/App.vue
    # App view now listens to TRANSACTION_ACTION notifications and
    # invokes BackendTransactionCache.forget() on each target transactions.
```

**Checklist**

### DIFF
--- a/front-end/src/renderer/App.vue
+++ b/front-end/src/renderer/App.vue
@@ -5,6 +5,7 @@ import { useRouter } from 'vue-router';
 import 'bootstrap/dist/js/bootstrap.bundle.min';
 
 import useUserStore from '@renderer/stores/storeUser';
+import useWebsocketSubscription from '@renderer/composables/useWebsocketSubscription.ts';
 
 import {
   provideDynamicLayout,
@@ -19,8 +20,10 @@ import OrganizationStatusModal from '@renderer/components/Organization/Organizat
 import GlobalModalLoader from '@renderer/components/GlobalModalLoader.vue';
 import GlobalAppProcesses from '@renderer/components/GlobalAppProcesses';
 import { ToastManager } from './utils/ToastManager';
-import { createLogger, getErrorMessage } from '@renderer/utils';
+import { createLogger, getErrorMessage, isLoggedInOrganization } from '@renderer/utils';
 import { AppCache } from './caches/AppCache';
+import { parseTransactionActionPayload } from '@renderer/utils/parseTransactionActionPayload.ts';
+import { TRANSACTION_ACTION } from '@shared/constants';
 
 /* Composables */
 const router = useRouter();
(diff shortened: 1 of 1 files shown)


svienot (approved):
LGTM

PR #2801, opened 2026-05-07
fix: enable wait on first connect for NATS JetStream service

### LINKED ISSUE #2775: Regression: signatures do not appear until relaunch on v0.29.0
**Description:**  
**User1** was using v0.27.0 and **User2** was using v0.29.0. In that setup, **User1** could see signatures from **User2**, but **User2** could not see signatures from **User1** until relaunching the app. After **User1** updated to v0.29.0, neither user could see the other person’s signatures until relaunch. This points to a regression in live signature updates between v0.27.0 and v0.29.0.

**Expected behavior:**  
When one user signs a transaction, the other user should see the signature update immediately without needing to relaunch.

**Actual behavior:**  
With **User1** on v0.27.0 and **User2** on v0.29.0, signature updates were only visible in one direction. After **User1** updated to v0.29.0, signature updates were no longer visible in either direction until relaunch.

**Notes:**  
- The issue should be investigated across v0.27.0, v0.28.0, and v0.29.0.
- Existing tests should be updated, or a new test should be added, to cover this behavior and prevent regressions in the future.
- Both notification pushes as well as selecting the affected transactions resulted in no updates.

jbair06:
**Description**:
This pull request introduces a configuration improvement to the NATS JetStream service. The main change ensures that the service will wait for the first connection to be established before proceeding, which can help prevent race conditions or errors on startup. 

If the connection is not established within 30 seconds, the service will continue to attempt to connect in the background, and will not prevent the pod from starting up.

**Related issue(s)**:

Fixes #2775 

**Checklist**

### DIFF
--- a/back-end/libs/common/src/nats/nats-jetstream.service.ts
+++ b/back-end/libs/common/src/nats/nats-jetstream.service.ts
@@ -60,6 +60,7 @@ export class NatsJetStreamService implements OnModuleDestroy {
       name: 'nestjs-service',
       maxReconnectAttempts: -1,
       reconnectTimeWait: 1000,
+      waitOnFirstConnect: true,
     });
 
     this.js = this.nc.jetstream();


ericleponner (approved):
LGTM

PR #2831, opened 2026-05-12
fix: organization transaction life cycle tests

### LINKED ISSUE #2804: test: re-enabled skipped tests from organizationTransactionLifecycleTests.test.ts
### Description

Two automation tests from `organizationTransactionLifecycleTests.test.ts` have been skipped by PR #2783.
Investigations are needed to understand why those tests fail and re-enable them.

### Steps to reproduce

Remove `skip` keywords from `organizationTransactionLifecycleTests.test.ts` and run corresponding tests.

```
  test.skip('Verify transaction is shown "Ready for Execution" and correct stage is displayed', async () => {
  test.skip('Verify transaction is shown "History" after it is executed', async () => {
```

### Additional context

_No response_

### Hedera network

_No response_

### Version

v0.29.0

### Operating system

None

yasenltd:
Summary

  - Re-enables the two previously-skipped tests in organizationTransactionLifecycleTests.test.ts ("Ready for Execution" and "History").
  - Replaces the API-based co-signing helper (signTransactionByAllUsersViaApi) with the original UI flow — log out, log in as each org user via logInAndSignTransactionByAllUsers, then log back in as the first user.

  Why

  After #2783 these two tests were temporarily skipped because the API-signing path was unstable for the AccountUpdate against a complex-key account: the transaction would stall in "Collecting Signatures" and never advance to "Ready
  for Execution"/"History", so the stage assertions timed out.

  Suspected front-end bug (why we stay on the UI path for now)

  We believe the API-signing failure is not in the automation helper but in the desktop app's handling of API-uploaded signatures for transactions whose key tree contains a nested threshold (the fixture key is 1 OF (user0_key, 2 OF 2
   (user1_key, user2_key))):

  - When users[1] and users[2] upload their signatures via POST /transactions/:id/signatures, the backend stores both TransactionSigner rows, but the FE's signature-progress evaluator appears to not recognize that the inner 2 OF 2
  threshold is satisfied — so the UI never marks the root threshold as satisfied and the executor never picks the transaction up.
  - The same multi-sig key works correctly when each user signs through the Ready to Sign UI flow: each click pushes the signature into the SDK transaction bytes via the same code path the executor reads, and the threshold evaluator
  agrees the key is satisfied. That's why the UI flow is reliable while the API flow is not.
  - Likely culprits to investigate in a follow-up:
    - the signature-aggregation/threshold-evaluation logic in the renderer (front-end/src/renderer/.../transactionSignatureProgress* or equivalent) that maps stored signer rows back onto the complex-key tree.
    - whether sdkTransaction.addSignature is being invoked for API-uploaded signatures on the same code path as UI-driven sign clicks, or whether the API path skips the SDK merge and only persists the signer row.

  This PR is intentionally a test-side workaround so the lifecycle coverage stops being skipped. A follow-up issue should be opened to fix the FE so signTransactionByAllUsersViaApi can be reintroduced and the UI flow (which costs
  ~30s per signer in extra logout/login) can be retired.

  Changes

  automation/tests/organization-advanced/organizationTransactionLifecycleTests.test.ts
  - test.skip(...) → test(...) for both lifecycle tests.
  - Co-signing block reverted to:
  await transactionPage.clickOnTransactionsMenuButton();
  await organizationPage.logoutFromOrganization();
  await organizationPage.logInAndSignTransactionByAllUsers(
    globalCredentials.password,
    txId ?? '',
  );
  await organizationPage.signInOrganization(
    firstUser.email,
    firstUser.password,
    globalCredentials.password,
  );
  - Kept the waitForStageCompleted(N) polling assertions from #2783.

  Test plan
  
  - CI green on Transaction tool project for @organization-advanced.
  - Local run (Solo localnet): both tests pass end-to-end.
    - Ready-for-Execution test reaches stage 1 completed, stage 2 not completed.
    - History test reaches all 4 stages completed with status SUCCESS.
  - No regression in the other three tests in the same describe (next-button visibility / redirect).

  Follow-up

  - File a FE issue to investigate why API-uploaded signatures don't satisfy nested-threshold complex keys; once fixed, revert this PR to use signTransactionByAllUsersViaApi and remove the per-signer logout/login from the lifecycle
  tests.


ericleponner (approved):
LGTM

PR #2832, opened 2026-05-12
fix: inconsistent visibility of save complex key

### LINKED ISSUE #2681: Inconsistent visibility of “Save Complex Key” button when re-adding existing public keys
### Description

The “Save Complex Key” button behaves inconsistently when a user reopens an existing Complex Key and re-adds previously used public keys. The button is displayed in scenarios where saving is not valid and would result in an error, creating confusion and a misleading user experience.

### Steps to reproduce

1. Create a new Complex Key with at least 2 public keys
2. Close the Complex Key screen
3. Reopen the Complex Key screen
4. Add the first already used public key
5. Observe that the “Save Complex Key” button is visible (unexpected, as only one key is present)
6. Add the second already used public key
7. Observe that the “Save Complex Key” button is still visible (unexpected, saving would result in an error)
8. Notice that the “Save Complex Key” button disappears afterward
9. Click the “Done” button
10. The existing Complex Key is assigned

Expected Result:

The “Save Complex Key” button should only be visible when the current state is valid for saving a new Complex Key
If the selected keys match an existing Complex Key, the UI should prevent saving and clearly indicate that the existing key will be reused

Actual Result:

The “Save Complex Key” button appears in invalid states where saving is not allowed
The button visibility changes inconsistently, leading to confusion
Attempting to save in these states would result in an error

Impact:

Poor user experience due to misleading UI behavior
Potential for user errors and unnecessary error messages

### Additional context

_No response_

### Hedera network

testnet

### Version

0.27.0

### Operating system

macOS

Kolpic:
Fixes: #2681

## Summary

The "Save Complex Key" button was visible when a user rebuilt an
already-saved complex key. Clicking it produced a "Complex key already
exists" error from the backend.

## What changed

- `KeyField.vue` pre-fetches the user's saved complex keys on mount and
  passes them down to `ComplexKeyModal.vue`.
- The modal hides "Save Complex Key" when the current keys match a
  saved entry, and shows a "Complex key already exists" warning next
  to "View Summary".
- In-place edits of an existing complex key keep the local cache in
  sync. On IPC failure, the cache and selected key are rolled back and
  an error toast is shown.
- Concurrent edits are ignored while a save is in flight to prevent
  cache corruption.
- The cache is refreshed when the "Add Existing" picker closes, so
  deletions there are reflected immediately.

## Tests

Added two new spec files. 

- `ComplexKeyModal.spec.ts`: button visibility matrix,
  warning span content, reactivity to inner key changes.
- `KeyField.spec.ts`: pre-fetch on mount, prop pass-through,
  synchronous match lookup, refresh after save, in-place edit
  success/rejection/concurrency, defensive rollback.

## Test plan

- [ ] Create a Complex Key with at least two public keys, save it,
      close.
- [ ] Add a second new key. "Save Complex Key" appears.
- [ ] Replace those with the two existing keys from step 1. "Save
      Complex Key" disappears and a warning shows the saved key's
      nickname.
- [ ] Remove one of the existing keys. The warning disappears.
- [ ] Click Done. The existing key is assigned to the field.

### DIFF
--- a/front-end/src/renderer/components/ComplexKey/ComplexKeyModal.vue
+++ b/front-end/src/renderer/components/ComplexKey/ComplexKeyModal.vue
@@ -49,10 +49,6 @@ const matchingSavedKey = computed<SavedComplexKey | null>(() => {
   return props.savedComplexKeys.find(k => k.protobufEncoded === encoded) ?? null;
 });
 
-const isSingleKey = computed(
-  () => currentKey.value instanceof KeyList && currentKey.value.toArray().length < 2,
-);
-
 /* Handlers */
 const handleShowUpdate = (show: boolean) => emit('update:show', show);
 
@@ -120,7 +116,7 @@ const modalContentContainerStyle = { padding: '0 10%', height: '80%' };
               >{{ summaryMode ? 'Edit Mode' : 'View Summary' }}</AppButton
             >
             <AppButton
-              v-if="onSaveComplexKey && !currentKeyInvalid && !matchingSavedKey && !isSingleKey"
+              v-if="onSaveComplexKey && !currentKeyInvalid && !matchingSavedKey"
               type="button"
               color="primary"
               class="ms-3"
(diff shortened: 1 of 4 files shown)


ericleponner (approved):
LGTM

PR #2839, opened 2026-05-12
chore(deps): Dependency security updates

jbair06:
**Description**:
This pull request updates several key dependencies to their latest versions and adds new overrides for improved security, compatibility, and bug fixes. 

**Notes**:
ajv is flagged as having a vulnerability. The only use that is not covered by the override is strictly for the electron-builders build process. It does not affect the final asset.


package.json
```diff
@@ -6,10 +6,10 @@
   },
   "pnpm": {
     "overrides": {
-      "protobufjs@<8.0.0": "7.5.5",
-      "protobufjs@>=8.0.0": "8.0.1",
+      "protobufjs@<8.0.0": "7.5.8",
+      "protobufjs@>=8.0.0": "8.2.0",
       "follow-redirects": "1.16.0",
-      "lodash": "4.17.21",
+      "lodash": "4.18.1",
```
steven-sheehy, line 12: We also have lodash in automation that needs to be bumped

jbair06: It should already be on 4.18.1. On further inspection, it isn't needed in automation. The one use is unnecessary. I'll remove it from automation and keep this override.

steven-sheehy (approved):

PR #2857, opened 2026-05-15
feat: allow keys in Public Keys to be used in Account transactions

### LINKED ISSUE #1578: Allow keys in Public Keys to be used in Account transactions
### Problem

The user is able to upload public keys, keys that the user does not own. Users will want to create or alter account keys using these public keys, and cannot current do so.

### Solution

The key creation or altering needs to show all public keys the user has access to, whether they are uploaded or owned by the user or contact. 
Add a new tab to this form for 'Public Keys' which will show all the public keys in the user has uploaded.

Image

### Alternatives

_No response_

ericleponner:
**Description**:
Changes below add an extra tab in `Key Chooser` modal which enables to select a key defined in `Settings` > `Public Keys`.

Before:
image

After:
image

**Related issue(s)**:

Fixes #1578

**Notes for reviewer**:

```
M   front-end/src/renderer/components/ComplexKey/ComplexKeyAddPublicKeyModal.vue
    # Added tab 'My public keys'.
    # Keys are fetched from storeUser.publicKeyMappings.
    
M   front-end/src/renderer/stores/storeUser.ts
    # Added a call to to refetchPublicKeys() to make sure that key mappings are loaded.
```

**Checklist**

### DIFF
--- a/front-end/src/renderer/components/ComplexKey/ComplexKeyAddPublicKeyModal.vue
+++ b/front-end/src/renderer/components/ComplexKey/ComplexKeyAddPublicKeyModal.vue
@@ -20,6 +20,7 @@ const logger = createLogger('renderer.component.complexKeyAddPublicKeyModal');
 /* Enums */
 enum KeyTab {
   MY = 'My keys',
+  PUBLIC_KEY = 'My public keys',
   CONTACTS = 'My contacts',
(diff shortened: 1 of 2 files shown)


svienot (approved):
LGTM

Issue #2895, opened 2026-05-27
Improving input validation in registered node create/update/delete


ericleponner:
### Problem

When creating, updating or deleting registered nodes, Create Draft form makes sure that user input is valid and enables / disables create command accordingly.
However this process is mostly silent . If registered node id is not found on chain, create command is disabled without any feedback.

### Solution

Adding some visual feedback next to input text field.
For example, when value in `Registered Node Id` does not match an existing node, we could display a message next to the input field.

One option could be to use the right part of the title line (see below).

<img width="638" height="109" alt="Image" src="https://github.com/user-attachments/assets/97562520-05cb-4a34-9f69-7a5222a80643" />

PR #2910, opened 2026-06-03
feat: improving service endpoint layout in Registered Node create/update form

ericleponner:
**Description**:

Changes below improve `Service Endpoint` editing in `Registered Node Create/Update` transaction draft.

Flag `Use domain name` is removed and `IP Address/Domain` textfield accepts both syntax (like in `Node Create`/`Update`):

- Before
image

- After
image

`Type`, `IP / Domain`, `Port` and `TLS` fields are now layed out on the same line:

- Before
image

- After
image

**Related issue(s)**:

Part 3 for #2895.

**Notes for reviewer**:

```
M   front-end/src/renderer/components/Transaction/Create/RegisteredNodeCreate/EndpointRow.vue
    # Now removes Use domain name checkbox and uses IpDomainInput which accepts 
    # both IP address and domain name in the same textfield.
    # Type, IP / Domain, Port and TLS flags are now placed on the same line.
    
A   front-end/src/renderer/components/IpDomainInput.vue
A   front-end/src/renderer/components/IpDomainInputStatus.ts
    # An <input> element setup to type either an IP address V4 or a domaine name.
    # IpDomainInput.props.value represents IP address as an Uint8Array and domain name as a string.
    # IpDomainINput.emit.status provides a status: empty, invalid and valid.
```

**Checklist**

### DIFF
--- a/front-end/src/renderer/components/IpDomainInput.vue
+++ b/front-end/src/renderer/components/IpDomainInput.vue
@@ -13,7 +13,7 @@ const emit = defineEmits<{
 const inputText = ref<string>('');
 
 /* Computed */
-const inputValue = computed<string | null>(() => {
+const inputValue = computed<string | Uint8Array | null>(() => {
   let result: string | Uint8Array | null;
 
   const trimmed = inputText.value.trim();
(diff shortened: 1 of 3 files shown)


svienot (approved):
LGTM

PR #2914, opened 2026-06-05
feat: automate the signing of the release artifacts and publish

### LINKED ISSUE #1220: Automate the signing of the release artifacts
### Problem

The builds are manually signed off. 

### Solution

Automate the signing of the builds. 

### Alternatives

_No response_

### LINKED ISSUE #2922: chore: update package versions to v{{VERSION}}
**Description:**
Automate snapshot version bumps so the PR title conforms to the required `chore: update package versions to v{{VERSION}}` format without manual intervention.

**Acceptance Tests:**
- Automated process bumps version and opens a PR with the exact title `chore: update package versions to v<new-version>`
- PR passes title format checks on creation

**Notes:**
Replace `{{VERSION}}` with the target version token used by your versioning toolchain.

jbair06:
**Description**:
This pull request significantly refactors the release automation workflow to support more flexible release modes, improve maintainability, and consolidate Docker image publishing. The workflow is split into modular jobs for preparing release metadata, cutting the release, building Docker images, building macOS artifacts, publishing the GitHub release, and creating snapshot bump PRs. The old standalone Docker image workflow is removed, and the release automation now supports three modes: full release, Docker images only, and macOS build only.

**Related issue(s)**:

Fixes #1220
Fixes #2922

**Notes for reviewer**:
Tested all paths on a forked repo.

**Checklist**


.github/workflows/release-automation.yaml
```diff
@@ -167,26 +477,39 @@ jobs:
...
+        with:
+          name: macos-artifacts-${{ needs.prepare.outputs.version }}
+          path: macos-artifacts
+
       - name: Create GitHub Release
         uses: step-security/release-action@a0d8e0f90f554c15366ca2c9046bf4090d24adbe # v1.21.0
         with:
           body: ${{ env.RELEASE_BODY }}
           commit: ${{ env.RELEASE_BRANCH }}
           draft: true
```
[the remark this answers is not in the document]
jbair06: We must have a draft, for now. The name of the job could be changed, perhaps. But we'll leave it as is for now.

.github/workflows/release-automation.yaml
```diff
@@ -130,27 +172,295 @@ jobs:
...
+        run: pnpm run build && pnpm exec electron-builder --mac --publish=never
+
+      - name: Prepare App Store Connect API Key
+        # notarytool needs the App Store Connect API key as a .p8 file on disk;
+        # the secret holds it base64-encoded. Write it to the runner temp dir
+        # and export its path for the notarization step.
+        run: |
+          API_KEY_PATH="${RUNNER_TEMP}/apple_api_key.p8"
+          echo "${{ secrets.APPLE_API_KEY_BASE64 }}" | base64 --decode > "${API_KEY_PATH}"
+          echo "APPLE_API_KEY=${API_KEY_PATH}" >> "$GITHUB_ENV"
```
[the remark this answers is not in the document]
jbair06: The finding is technically correct about portable macOS systems, but GitHub Actions macOS runners ship with GNU coreutils in the PATH ahead of the BSD tools, so --decode works there. 

 It's a valid portability observation if this script ever ran on a stock macOS machine, but for CI it's a non-issue.

.github/workflows/release-automation.yaml
```diff
@@ -130,27 +172,295 @@ jobs:
...
+          INSTALLERS=(
+            release/*-"${{ env.VERSION }}"-mac-*.dmg
+            release/*-"${{ env.VERSION }}"-mac-*.pkg
+          )
+          for FILE in "${INSTALLERS[@]}"; do
+            xcrun notarytool submit "${FILE}" "${NOTARY_ARGS[@]}" --wait
+          done
+          for FILE in "${INSTALLERS[@]}"; do
+            xcrun stapler staple "${FILE}"
+          done
```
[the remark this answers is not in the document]
jbair06: fixed

front-end/scripts/load-release-env.sh
```diff
@@ -0,0 +1,34 @@
...
+    line="${line#"${line%%[![:space:]]*}"}"    # strip leading whitespace
+    [[ -z "${line}" || "${line}" == \#* ]] && continue
+    [[ "${line}" != *=* ]] && continue
+    key="${line%%=*}"
+    val="${line#*=}"
+    key="${key//[[:space:]]/}"                 # env names have no spaces
+    val="${val%\"}"; val="${val#\"}"           # strip one pair of surrounding quotes
+    val="${val%\'}"; val="${val#\'}"
+    [[ -n "${!key:-}" ]] && continue           # don't clobber the existing env (CI/shell wins)
+    export "${key}=${val}"
```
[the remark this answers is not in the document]
jbair06: unlikely to be an issue, but fixed.

.github/workflows/release-automation.yaml
```diff
@@ -8,6 +8,13 @@ on:
       version:
         description: "Version (semver)"
         required: true
+      mode:
+        description: "Pipeline scope"
+        type: choice
+        options:
+          - full_release
+          - publish_docker_images_only
+          - test_mac_build_only
```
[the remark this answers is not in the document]
jbair06: type: choice sets the first item as default, it appears. This is unnecessary

.github/workflows/release-automation.yaml
```diff
@@ -130,27 +172,299 @@ jobs:
...
+        run: pnpm run build && pnpm exec electron-builder --mac --publish=never
+
+      - name: Prepare App Store Connect API Key
+        # notarytool needs the App Store Connect API key as a .p8 file on disk;
+        # the secret holds it base64-encoded. Write it to the runner temp dir
+        # and export its path for the notarization step.
+        run: |
+          API_KEY_PATH="${RUNNER_TEMP}/apple_api_key.p8"
+          echo "${{ secrets.APPLE_API_KEY_BASE64 }}" | base64 --decode > "${API_KEY_PATH}"
+          echo "APPLE_API_KEY=${API_KEY_PATH}" >> "$GITHUB_ENV"
```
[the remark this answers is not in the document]
jbair06: My understanding is this is not actually an issue.

front-end/scripts/load-release-env.sh
```diff
@@ -0,0 +1,35 @@
...
+    [[ -z "${line}" || "${line}" == \#* ]] && continue
+    [[ "${line}" != *=* ]] && continue
+    key="${line%%=*}"
+    val="${line#*=}"
+    key="${key//[[:space:]]/}"                 # env names have no spaces
+    [[ -z "${key}" ]] && continue              # skip lines with no key (e.g. =value)
+    val="${val%\"}"; val="${val#\"}"           # strip one pair of surrounding quotes
+    val="${val%\'}"; val="${val#\'}"
+    [[ -n "${!key:-}" ]] && continue           # don't clobber the existing env (CI/shell wins)
+    export "${key}=${val}"
```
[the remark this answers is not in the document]
jbair06: fixed

.github/workflows/smoke-test-mac.yaml
```diff
@@ -0,0 +1,122 @@
+# SPDX-License-Identifier: Apache-2.0
+
+name: macOS Smoke Test
```
steven-sheehy, line 3: Is this something we're going to run every release? Does it need to be added to the checklist? Or this is only when we need to update some secrets.

.github/workflows/smoke-test-mac.yaml
```diff
@@ -0,0 +1,122 @@
...
+          )
+          if [[ ${#INSTALLERS[@]} -eq 0 ]]; then
+            echo "No .dmg or .pkg installers found — electron-builder did not produce expected artifacts" >&2
+            exit 1
+          fi
+          for FILE in "${INSTALLERS[@]}"; do
+            xcrun notarytool submit "${FILE}" "${NOTARY_ARGS[@]}" --wait
+          done
+          for FILE in "${INSTALLERS[@]}"; do
+            xcrun stapler staple "${FILE}"
```
steven-sheehy: Why can't we reuse the shell script here?

.github/workflows/release-automation.yaml
```diff
@@ -130,27 +163,279 @@ jobs:
...
+          )
+          if [[ ${#INSTALLERS[@]} -eq 0 ]]; then
+            echo "No .dmg or .pkg installers found — electron-builder did not produce expected artifacts" >&2
+            exit 1
+          fi
+          for FILE in "${INSTALLERS[@]}"; do
+            xcrun notarytool submit "${FILE}" "${NOTARY_ARGS[@]}" --wait
+          done
+          for FILE in "${INSTALLERS[@]}"; do
+            xcrun stapler staple "${FILE}"
```
steven-sheehy: What's the point of notarize-mac-installer.sh if we don't reuse it here? Should delete it or use it here.

.github/workflows/release-automation.yaml
```diff
@@ -130,27 +163,279 @@ jobs:
...
+        # artifacts are uploaded via upload-artifact.
+        env:
+          # Developer ID Application cert (signs the .app, .dmg and .zip).
+          # CSC_LINK accepts the base64-encoded .p12 directly.
+          CSC_LINK: ${{ secrets.MACOS_CERT_APP_BASE64 }}
+          CSC_KEY_PASSWORD: ${{ secrets.MACOS_CERT_APP_PASSWORD }}
+          # Developer ID Installer cert (signs the .pkg).
+          CSC_INSTALLER_LINK: ${{ secrets.MACOS_CERT_INSTALLER_BASE64 }}
+          CSC_INSTALLER_KEY_PASSWORD: ${{ secrets.MACOS_CERT_INSTALLER_PASSWORD }}
+        run: pnpm run build && pnpm exec electron-builder --mac --publish=never
```
steven-sheehy, line 352: What's the point of build-mac.sh if we don't use it here?

.github/workflows/release-automation.yaml
```diff
@@ -130,27 +163,279 @@ jobs:
...
+          images: ${{ env.REGISTRY }}/${{ matrix.image }}
+          tags: type=raw,value=${{ env.VERSION }}
+
+      - name: Build and Push Docker Image
+        id: push
+        uses: docker/build-push-action@f9f3042f7e2789586610d6e8b85c8f03e5195baf # v7.2.0
+        with:
+          context: .
+          file: ${{ matrix.dockerfile }}
+          platforms: linux/amd64
```
steven-sheehy, line 251: Why not also produce linux/arm64 for use running locally?

.github/workflows/smoke-test-mac.yaml
jbair06: Not every release, no. This is more of a "break glass" workflow, when something's broken or changed: rotating certs/secrets, an Apple agreement update, or tweaking electron-builder config and wanting to validate before cutting a real release.

.github/workflows/smoke-test-mac.yaml
jbair06: We should be, I've updated it.

.github/workflows/release-automation.yaml
jbair06: We should be using it here. The script was designed to work both in a workflow and locally. It should be used it. I fixed it.

.github/workflows/release-automation.yaml
jbair06: The point is for local building, to use local env variables for certs and passwords. I was using the keychain to store the certs, but wanted to better mirror how it is to be done in CI/CD, and make the setup process a bit easier for other developers to build the app locally.

jbair06: I suppose we don't need both the local building option AND the smoke-test-mac job. My thinking was that if a user only has access to certs, or only has access to the actions, they can do whichever they are able to do. Maybe overkill?

.github/workflows/release-automation.yaml
jbair06: There isn't currently a need for it. Currently, we all build the backend locally for testing. That said, we could add the arm64 option, just increases build time a bit. 

It also could be good to add arm64 if we plan to use different machines in our deployments. I'll leave it unless you want both.

steven-sheehy: The image build will be done in parallel with architectures. We should add it as it's useful for debugging production issues of the same version locally and other users may find it useful.

.github/workflows/release-automation.yaml
steven-sheehy: YAGNI. Should remove it if you have the smoke test.

front-end/scripts/release.env.example
steven-sheehy, line 1: Think this file could be removed and its format discussed inside the RELEASE.md instead. Should create a config properties table with name and description.

jbair06: Done.

.github/workflows/release-automation.yaml
jbair06: Done. And all related code.

.github/workflows/release-automation.yaml
jbair06: Done. It requires docker/setup-qemu-action in order to build for arm64 on an amd64 runner.

steven-sheehy (dismissed):

steven-sheehy (dismissed):

steven-sheehy (approved):

PR #3004, opened 2026-06-17
feat: display file contents in modal instead of saving to disk

### LINKED ISSUE #2977: Show file contents in modal instead of saving to disk in FileDetails
**Description:**
In `FileDetails.vue`, the "View" link on `transaction.contents` calls `saveFile()`, opening an OS save dialog. The label implies display, not export. Reviewers signing a transaction need to read the contents in-app.

**Acceptance Tests:**
- Clicking "View" on a `FileUpdateTransaction` opens a modal; content is decoded via `decodeProto` if the file ID is a special file ID
- Clicking "View" on a `FileCreateTransaction` shows decoded text if readable, otherwise shows a "binary content — cannot display" message
- Content exceeding a reasonable size limit skips the modal and shows the size with a note that it is too large to display
- `saveFile` is fully removed from `FileDetails.vue` (import and call)
- Modal renders correctly in both `TransactionDetails.vue` and `TransactionBrowserPage.vue`

**Notes:**
Key files:
- `FileDetails.vue:255` — the "View" `<span>` and its `@click` handler
- `KeyStructureModal.vue` — existing modal pattern in the same component to follow
- `electronUtilsService.ts` — `saveFile` can be removed from imports here

jbair06:
**Description**:
This pull request introduces a new modal component for viewing file contents in the UI, improves the user experience when displaying Hedera file contents, and adds comprehensive tests for the new functionality. The most important changes are summarized below.

**Related issue(s)**:

Fixes #2977 

**Checklist**

### DIFF
--- a/front-end/src/renderer/components/FileContentsModal.vue
+++ b/front-end/src/renderer/components/FileContentsModal.vue
@@ -1,0 +1,51 @@
+<script setup lang="ts">
+import { computed } from 'vue';
+import AppModal from '@renderer/components/ui/AppModal.vue';
+import { isHederaSpecialFileId, decodeProto } from '@shared/hederaSpecialFiles';
+
+const props = defineProps<{
+  show: boolean;
+  contents: Uint8Array;
+  fileId?: string | null;
+}>();
+
+const emit = defineEmits(['update:show']);
+
+const handleShowUpdate = (show: boolean) => emit('update:show', show);
+
+const displayContent = computed<string | null>(() => {
+  if (!props.contents?.length) return '';
+
+  if (props.fileId && isHederaSpecialFileId(props.fileId)) {
+    try {
+      return decodeProto(props.fileId, props.contents) ?? '';
+    } catch {
+      return null;
+    }
+  }
+
+  try {
+    const text = new TextDecoder('utf-8', { fatal: true }).decode(props.contents);
+    if (/[\x00-\x08\x0B\x0C\x0E-\x1F\x7F]/.test(text)) {
+      return null;
+    }
+    return text;
+  } catch {
+    return null;
+  }
+});
+</script>
+<template>
+  <AppModal :show="show" @update:show="handleShowUpdate" class="medium-modal">
+    <div class="p-5">
+      <h3 class="text-center text-title text-bold mb-4">File Contents</h3>
+      <pre
+        v-if="displayContent !== null"
+        class="m-0 overflow-auto text-small"
+        style="max-height: 60vh; white-space: pre-wrap; word-break: break-all"
+        >{{ displayContent || '(empty)' }}</pre
+      >
+      <p v-else class="m-0 text-small">Binary content — cannot display</p>
+    </div>
+  </AppModal>
+</template>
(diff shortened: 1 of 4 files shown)


front-end/src/renderer/components/FileContentsModal.vue
```diff
@@ -0,0 +1,51 @@
...
+    return text;
+  } catch {
+    return null;
+  }
+});
+</script>
+<template>
+  <AppModal :show="show" @update:show="handleShowUpdate" class="medium-modal">
+    <div class="p-5">
+      <h3 class="text-center text-title text-bold mb-4">File Contents</h3>
```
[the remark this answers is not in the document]
jbair06: fixed

front-end/src/renderer/components/Transaction/Details/FileDetails.vue
```diff
@@ -265,5 +289,12 @@ const commonColClass = 'col-6 col-lg-5 col-xl-4 col-xxl-3 overflow-hidden py-3';
       v-model:show="isKeyStructureModalShown"
       :account-key="new KeyList(transaction.keys)"
     />
+
+    <FileContentsModal
+      v-if="transaction.contents !== null"
+      v-model:show="isFileContentsModalShown"
+      :contents="transaction.contents"
+      :file-id="contentsFileId"
+    />
```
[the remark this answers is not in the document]
jbair06: fixed

front-end/src/renderer/components/FileContentsModal.vue
```diff
@@ -0,0 +1,51 @@
...
+    }
+    return text;
+  } catch {
+    return null;
+  }
+});
+</script>
+<template>
+  <AppModal :show="show" @update:show="handleShowUpdate" class="medium-modal">
+    <div v-if="show" class="p-5">
```
svienot, line 44: This new dialog works fine and is consistent with the _See details_ of the _Key_ field on a File Create transaction.
However, for those 2 modals, I notice we do not show a top-left cross to close the dialog (equivalent to clicking outside) like we do in most (all ?) of the other dialogs in the tool. I am wondering whether we should have one cross here (and in the Key Details dialog) for the sake of overall consistency...

front-end/src/renderer/components/FileContentsModal.vue
```diff
@@ -0,0 +1,51 @@
...
+      <h3 class="text-center text-title text-bold mb-4">File Contents</h3>
```
svienot: Minor remark: the title of the dialog is _File Contents_, which in the case of a File Append transaction might be slightly misleading since in this case we do not show what the resulting File Contents will be.

front-end/src/renderer/components/FileContentsModal.vue
jbair06: I added the `x` to both this modal and KeyStructureModal

front-end/src/renderer/components/FileContentsModal.vue
jbair06: The title now says `File Contents` or `Append Contents`

svienot (approved):
LGTM

PR #3021, opened 2026-06-24
fix: registered node update transactions show service endpoints with commas

### LINKED ISSUE #3014: Registered node update transactions show service endpoints with commas instead of ./: in octets?
### Description

The Registered Node Update transaction display page seems to be improperly displaying IPv4 endpoints, showing commas instead of dots.

It seems like this might be an issue in RegisteredNodeDetails.vue; the JS SDK returns IPv4 addresses as a list of octets, and it's just by default creating a comma-separated list instead of specifically joining with a "." like the other display pages do. (For example, NodeDetails.vue uses another getter that returns it as a string.)

### Steps to reproduce

1. Generate a Registered Node Update with an IPV4 address
2. Observe that it looks like 1,1,1,1 instead of 1.1.1.1

### Additional context

_No response_

### Hedera network

_No response_

### Version

v0.32.0

### Operating system

macOS

svienot:
**Description**:

Fixes: #3014 

When displaying the details of a `RegisteredNodeCreate` or `RegisteredNodeUpdate` transaction, the `RegisteredNodeDetails` view needs to treat `endpoint.ipAddress` as `Uint8Array` and not as a `string` in order to display `1.2.3.4 `instead of `1,2,3,4.`

Add some unit tests for RegisteredNodeDetails.vue

### DIFF
--- a/front-end/src/renderer/utils/sdk/getData.ts
+++ b/front-end/src/renderer/utils/sdk/getData.ts
@@ -314,6 +314,10 @@ const getRegisteredEndpointType = (
   return endpoint.type;
 };
 
+export function stringifyIpAddressBytes(ipBytes: Uint8Array | null | undefined): string {
+  return ipBytes && ipBytes.length === 4 ? Array.from(ipBytes).join('.') : '';
+}
+
 /**
  * Extract an endpoint from a decoded SDK transaction back into the FE-facing
  * shape. Deliberately does NOT populate `uiId` — that field is a render-side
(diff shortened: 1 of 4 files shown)


front-end/src/renderer/utils/sdk/getData.ts
```diff
@@ -325,9 +329,7 @@ const getRegisteredEndpointType = (
 export const getComponentRegisteredEndpoint = (
   endpoint: RegisteredServiceEndpoint,
 ): ComponentRegisteredServiceEndpoint => {
-  const ipBytes = endpoint.ipAddress;
-  const ipAddressV4 =
-    ipBytes && ipBytes.length === 4 ? Array.from(ipBytes).join('.') : '';
+  const ipAddressV4 = stringifyIpAddressBytes(endpoint.ipAddress);
```
steven-sheehy, line 376: Can we ensure we support ipv6 as well here and render it properly? The ip_address field can support both.

svienot: Updated the `stringifyIpAddressBytes` routine to also handle IPv6 addresses.
Added unit tests for this function.

steven-sheehy (approved):

PR #3056, opened 2026-07-03
feat: revise email notification system to reduce per transaction noise

### LINKED ISSUE #2970: Revise email notification system to reduce per-transaction noise
**Description:**
The email system can fire individual notifications for each transaction in a group, which generates excessive volume — e.g. a group of 10 transactions staggered by seconds produces 10 separate emails per user per lifecycle event (created, signed, executed, expired). Notifications should be scoped to the group lifecycle or batched by time window.

**Acceptance Tests:**
- Transactions that belong to a group do not trigger per-transaction emails for events that have a group-level equivalent (created, ready to execute, executed, expired)
- Group lifecycle emails correctly reflect aggregate state (e.g. "all transactions signed, group ready to execute")
- Solo transactions (no group) continue to receive standard per-transaction notifications, unaffected
- A batching/digest option exists for high-frequency events (e.g. daily summary of executed transactions)

**Notes:**
Two viable approaches worth evaluating before implementation: (1) group-scoped notifications keyed to group lifecycle events, (2) time-windowed digests (hourly or daily) for certain event types. These are not mutually exclusive.

jbair06:
**Description**:
This pull request introduces support for new transaction notification types (`FAILED` and `REJECTED`) and implements robust group email batching for transaction status updates. The batching logic ensures that users receive a single consolidated email per group when all relevant transactions reach the same lifecycle stage, reducing email noise and improving clarity. Additionally, the email sending infrastructure is made more resilient by ensuring all queued emails are flushed on service shutdown, and several configuration and test improvements are included.

**Related issue(s)**:

Fixes #2970 

**Checklist**


back-end/libs/common/src/database/entities/notifications/notification.entity.ts
ericleponner, line 1: - Shouldn't we add `TRANSACTION_FAILED` and `TRANSACTION_REJECTED` to:
 `front-end/src/shared/interfaces/organization/notification-receiver/index.ts` ?
- Would it make sense to have these new notifications appear in `Settings` > `Notifications` next to other transaction notifications ?

ericleponner (dismissed):

back-end/libs/common/src/database/entities/notifications/notification.entity.ts
jbair06: 1. Yes, might as well. It won't matter, but when it is needed in the future, we might miss it if we don't do it now.
2. Not yet. The emails are not supported for those 2 types, so we don't want the user to see them in the settings yet.

ericleponner (approved):
LGTM

PR #3058, opened 2026-07-03
fix: defer password change to end of migration flow

### LINKED ISSUE #2994: Defer password change to end of migration flow
**Description:**
After a successful org connection, migration immediately triggers a password change. If the user exits before migration completes, their password is updated silently — causing every subsequent migration attempt to fail with invalid credentials.

**Acceptance Tests:**
- Password change does not occur until migration is fully complete
- Exiting migration after org connection but before completion leaves credentials unchanged
- A completed migration still results in the password being updated

ericleponner:
**Description**:
Changes below update `Migrate` and `SetupOrganization` views so that password change is performed before `Summary` step in migration flow.

`PerformSetup` is a new step  in `Migrate` wizard:
- it is inserted before `Summary` step
- it displays a progress message and has no user input or actions
- it triggers organization setup and password change that used to be done in `SetupOrganization`
- it appears very shortly since organization setup and password changes are usually fast
- `SetupOrganization` now only validates inputs and no longer makes organization related changes

image

**Related issue(s)**:

Fixes #2994

**Notes for reviewer**:

```
A   front-end/src/renderer/pages/Migrate/components/PerformSetup.vue
M   front-end/src/renderer/pages/Migrate/Migrate.vue
    # New tab inserted between SelectKeys and Summary tabs..
    # It triggers organization/password setup.
    # It also displays DecryptKeys modal when extracting keys.

M   front-end/src/renderer/pages/Migrate/components/SetupOrganization.vue
    # SetupOrganization now performs input validation only.
    # Organization setup and new/old password change are now done later in PerformSetup.vue

M   front-end/src/renderer/pages/Migrate/components/SelectKeys.vue
M   front-end/src/renderer/components/KeyPair/ImportEncrypted/components/EncryptedKeysBox.vue
    # Adjusted css in SelectKeys for visual consistency with other Migrate tabs

R   front-end/src/renderer/pages/Migrate/components/BeginKeysImport.vue
    # Logic from BeginKeysImport is now performed by PerformSetup view.
    # View is removed.
```

**Checklist**


front-end/src/renderer/pages/Migrate/Migrate.vue
```diff
@@ -112,24 +121,21 @@ const handleSetOrganizationId = async (value: string | null) => {
...
   }
 };
 
-const handleKeysImported = async (value: number) => {
+const didPerformSetup = async (importedKeyCount: number) => {
   if (!personalUser.value) throw new Error('(BUG) Personal User not set');
-  if (!value) {
+  if (importedKeyCount === 0) {
     await accountSetupStore.storeSkipRecoveryPhraseClaim();
   }
```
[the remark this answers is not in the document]
ericleponner: Fixed

front-end/src/renderer/pages/Migrate/Migrate.vue
```diff
@@ -219,20 +212,28 @@ const initializeUserStore = async () => {
...
 
         <!-- Summary Step -->
         <template v-if="stepIs('summary')">
-          <Summary :imported-keys-count="keysImported" :imported-user-data="importedUserData" />
+          <Summary
+            :imported-keys-count="keysImported"
+            :imported-user-data="importedUserData"
+            :personal-user="personalUser"
+            :organization-setup="organizationSetup"
+          />
```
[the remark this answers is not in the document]
ericleponner: Fixed

front-end/src/renderer/pages/Migrate/components/PerformSetup.vue
```diff
@@ -1,54 +1,117 @@
...
+  );
+
+  // 4) Add Organization Credentials
+  addOrganizationCredentials(
+    email,
+    setup.newOrganizationPassword,
+    organizationId,
+    props.personalUser.personalId,
+    jwtToken,
+    setup.newOrganizationPassword,
```
jbair06: This second password should be the app password, or keychain if selected, right? Looking at other usages, the personal password is used here.

front-end/src/renderer/pages/Migrate/components/PerformSetup.vue
```diff
@@ -1,54 +1,117 @@
...
+
+  // 3) Set new password
+  await changePassword(
+    setup.organizationURL,
+    setup.temporaryOrganizationPassword,
+    setup.newOrganizationPassword,
+  );
+
+  // 4) Add Organization Credentials
+  addOrganizationCredentials(
```
jbair06: I think we need to `await` this call so failures don't get swallowed or lost.

front-end/src/renderer/pages/Migrate/components/PerformSetup.vue
```diff
@@ -133,23 +163,68 @@ const restoreKeyPair = async (index: number, nickname: string, upload: boolean)
...
-    await decryptKeysRef.value?.process(encryptedKeyPaths, props.recoveryPhrase?.words);
+  for (let i = 0; i < userKeysWithMnemonic.length; i++) {
+    const userKey = userKeysWithMnemonic[i];
+
+    /**
+     * Restore the key if its mnemonic hash matches the currently imported recovery phrase.
+     * Do not restore if the key is already present in `user.keyPairs`.
+     * Multiple mnemonic phrases are supported; keys from other mnemonic phrases are not deleted.
+     */
+    if (userKeyHasMnemonic(userKey)) {
```
jbair06, line 176: The for loop is a filtered list of only keys with mnemonics, so this extra guard should be unnecessary.

front-end/src/renderer/pages/Migrate/components/PerformSetup.vue
```diff
@@ -1,54 +1,117 @@
...
+  if (user.organizations[0]) {
+    await user.selectOrganization(user.organizations[0]);
+  }
+
+  // 6) Wait a little to avoid flash effect and help user identify what's happening
+  await eyePersistence;
+};
+
+const startKeyImport = async () => {
+  if (props.selectedKeys && props.selectedKeys.length > 0) {
```
jbair06: `props.selectedKeys` cannot be undefined or null here, I think. And this function is only ever called if the list is > 0, so I think this guard is unnecessary, and the `else` path is never reached.

front-end/src/renderer/pages/Migrate/components/PerformSetup.vue
```diff
@@ -1,54 +1,117 @@
...
+    const selectedKeyPaths = props.selectedKeys.map(key => key.filepath);
+    await decryptKeysRef.value?.process(selectedKeyPaths, props.recoveryPhrase?.words);
```
jbair06: Can decryptKeysRef.value ever be null/undefined here? If it can, I believe this line would silently hang the migration process because it could not progress. Should probably just remove the `?`

front-end/src/renderer/pages/Migrate/Migrate.vue
```diff
@@ -112,24 +121,22 @@ const handleSetOrganizationId = async (value: string | null) => {
   if (allUserKeysToRecover.value.length !== 0) {
     step.value = 'selectKeys';
   } else {
-    step.value = 'summary';
+    step.value = 'performSetup';
   }
 };
 
-const handleKeysImported = async (value: number) => {
+const didPerformSetup = async (importedKeyCount: number) => {
```
jbair06: The `emit` from `PerformSetup` is declared with 2 parameters. Should either add the second one here, or remove from the declaration.

jbair06 (changes requested):

front-end/src/renderer/pages/Migrate/components/PerformSetup.vue
ericleponner: Fixed

front-end/src/renderer/pages/Migrate/components/PerformSetup.vue
ericleponner: Correct. It's a copy/paste from legacy code.
I updated the code.

front-end/src/renderer/pages/Migrate/components/PerformSetup.vue
ericleponner: Since `startKeyImport()` is called by `onMounted()`, `decryptKeysRef` should be resolved and not null.
? is required for type checking reason and protect the code.

front-end/src/renderer/pages/Migrate/components/PerformSetup.vue
ericleponner: `restoreExistingKeys()` is a copy of legacy [BeginKeysImport](https://github.com/hashgraph/hedera-transaction-tool/blob/05dd0d246b224dade5318e10a27c068b833f09d1/front-end/src/renderer/pages/Migrate/components/BeginKeysImport.vue#L64). And this part is pretty obscure to me. I would prefer to leave it untouched. But I can remove the check : let me know.

front-end/src/renderer/pages/Migrate/components/PerformSetup.vue
ericleponner: Good catch. Original code was indeed:
```
    addOrganizationCredentials(
      email,
      formData.newOrganizationPassword,
      organizationId.value,
      props.personalUser.personalId,
      organizationJwtToken.value || '',
      props.personalUser.password,
      true,
    ),
```
I replaced `setup.newOrganizationPassword` by `props.personalUser.password` (line 90).
Strangely I didn't see any effect during my tests (?).

front-end/src/renderer/pages/Migrate/Migrate.vue
ericleponner: Fixed

front-end/src/renderer/components/KeyPair/ImportEncrypted/components/EncryptedKeysBox.vue
```diff
@@ -59,7 +59,7 @@ watch(
         data-testid="checkbox-select-all-keys-component"
         class="cursor-pointer"
       />
-      <span>Select all</span>
+      <span class="text-secondary text-small">Select all</span>
     </div>
     <ul class="overflow-x-hidden" style="max-height: 30vh">
```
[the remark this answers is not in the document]
jbair06: This is unchanged from main, no need for a change at the moment.

front-end/src/renderer/pages/Migrate/Migrate.vue
```diff
@@ -219,20 +222,26 @@ const initializeUserStore = async () => {
...
+        <!-- Perform Migration Step -->
+        <template v-if="stepIs('performSetup')">
+          <PerformSetup
+            :personal-user="personalUser!"
+            :organization-setup="organizationSetup"
+            :recovery-phrase="recoveryPhrase"
+            :recovery-phrase-password="recoveryPhrasePassword"
+            :selected-keys="selectedKeysToRecover"
+            @didPerformSetup="didPerformSetup"
+          />
```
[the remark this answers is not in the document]
jbair06: They can run in parallel. This is not an issue.

jbair06 (approved):
LGTM

PR #3072, opened 2026-07-08
fix: skip signature validation when mirror node key resolution fails

### LINKED ISSUE #2930: Fix fail-open signature validation when mirror node key resolution fails
**Description:**
`hasValidSignatureKey` (`back-end/libs/common/src/utils/sdk/key.ts:31-46`) evaluates `0 >= (key.threshold || keys.length)` when given an empty `KeyList`, returning `true` — treating an unknown signing requirement as satisfied. The empty list originates in `TransactionSignatureService.computeSignatureKey` (`back-end/libs/common/src/transaction-signature/transaction-signature.service.ts:38-62`), where every `add*Key` helper silently swallows mirror-node fetch failures and pushes nothing. Any mirror hiccup during status evaluation causes under-signed transactions to advance to `WAITING_FOR_EXECUTION`, get submitted, be rejected with `INVALID_SIGNATURE`, and land in terminal `FAILED` — permanently consuming the TransactionId and forcing a full rebuild.

**Acceptance Tests:**
- `hasValidSignatureKey` returns `false` for an empty `KeyList` (zero threshold, zero children)
- Each `add*Key` helper propagates mirror fetch errors rather than catching and discarding them
- `processTransactionStatus` keeps a transaction in `WAITING_FOR_SIGNATURES` when required-key resolution throws, rather than advancing it
- `getValidatedSDKTransaction` (`back-end/libs/common/src/execute/execute.service.ts:219`) behaves the same way under key-resolution failure
- Unit tests cover: empty key list, partial mirror failure (some helpers succeed, one throws), and full mirror outage

**Notes:**
Consider emitting a metric/alert on required-key resolution failure rate so mirror degradation is visible before it manifests as failed governance transactions.

jbair06:
**Description**:
Fixes a fail-open vulnerability where mirror node fetch failures during signature key resolution caused `computeSignatureKey` to return an empty KeyList, which `hasValidSignatureKey` incorrectly evaluated as satisfied — allowing under-signed transactions to advance to `WAITING_FOR_EXECUTION` and be submitted with invalid signatures.                                                                    
                                                                                                                                                                                                                                                                                                                                                                                                              
Mirror failures now propagate out of `computeSignatureKey`. `processTransactionStatus` catches and skips affected transactions to preserve their current status, while the execution paths (`collateAndExecute`, `getValidatedSDKTransaction`) catch and proceed without signature validation — since mirror unavailability does not imply network unavailability. Also fixes a stale in-memory status bug in `prepareTransactions` and extends last-chance execution attempts to `WAITING_FOR_SIGNATURES` transactions whose `validStart` has become executable.    

**Related issue(s)**:

Fixes #2930 

**Checklist**


svienot (approved):
LGTM

PR #3073, opened 2026-07-08
fix: it is not possible to edit the associated registered nodes for a consensus node

### LINKED ISSUE #2981: It is not possible to edit the associated registered nodes for a consensus node
### Description

When creating a `Node Update Transaction`, there is no way for the user to view and edit (or delete) the current value of the `Associated Registered Nodes`. The only option is to blindly replace the field value with a new (set of) node(s).

### Steps to reproduce

Create New > Transaction > Node > Node Update
Enter the `node ID` of the node to update
This populates certain fields with their current value (such as `Node Account ID`, `Admin Key`, `Description`)

Issue 1/ The user would expect at this point to see the current (set of) associated registered node(s) to be able to alter it.
Issue 2/ The user has no way to remove existing associated registered node(s). Submitting with an empty list is a NOOP
Issue 3/ The Create and Share button is enabled even if no field value has been changed

### Additional context

_No response_

### Hedera network

other

### Version

0.33.0-SNAPSHOT

### Operating system

macOS

svienot:
**Description**:

- Fix typo in definition of `NetworkNode` interface (`associated_registered_node` -> `associated_registered_nodes`) and all its references in the code and tests. This was preventing the proper handling of this field in the creation of a `NodeUpdate` transaction, since the current value obtained from MN appeared as never defined.

- Add logic in `NodeUpdate` to detect when there is any change brought to the node properties and enable the _Share_/_Execute_ button accordingly.

**Related issue(s)**:

Fixes #2981


ericleponner (approved):
LGTM

PR #3076, opened 2026-07-08
feat: add additional metadata fields to transaction signer entity

### LINKED ISSUE #3071: Add version field to transaction_signer entity
**Description:**
There is no way to distinguish a user signing a transaction directly from a user importing signatures on behalf of other users who have accounts and keys on the system. Add a version (or similar) field to the transaction_signer entity to record which signing tool/version produced the signature (e.g. ttv1 v.14.8, ttv2 0.35.0).

**Acceptance Tests:**
- transaction_signer entity stores a version/tool field on creation
- Field is populated correctly for both direct signing and signature import flows
- Existing transaction_signer records without this field do not break on read

jbair06:
**Description**:
This pull request extends the transaction signature system to capture and persist additional metadata about the signing process, including the tool used, the recorder, and the frontend version. These enhancements are reflected throughout the DTOs, service logic, and tests to ensure the new fields are handled correctly and consistently.

**Related issue(s)**:

Fixes #3071 

**Checklist**

### DIFF
--- a/back-end/apps/api/src/transactions/dto/transaction-signer.dto.ts
+++ b/back-end/apps/api/src/transactions/dto/transaction-signer.dto.ts
@@ -13,6 +13,15 @@ export class TransactionSignerDto {
   userKeyId: number;
 
   @Expose()
+  recorderId: number | null;
+
+  @Expose()
+  tool: string | null;
+
+  @Expose()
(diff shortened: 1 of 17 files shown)


svienot (approved):
LGTM

PR #3093, opened 2026-07-08
fix: code scanning alert no. 7 and 8 - insecure randomness

jbair06:
Potential fix for [https://github.com/hashgraph/hedera-transaction-tool/security/code-scanning/7](https://github.com/hashgraph/hedera-transaction-tool/security/code-scanning/7)

Use Node.js cryptographic randomness for password generation. In this file, the best fix is to replace letter generation logic that currently uses `Math.random()` with `crypto.randomInt(26)` (or equivalent secure byte logic). This keeps the same output format (`xxxxx-xxxxx`, lowercase letters) and avoids functionality changes while removing predictability.

Concretely in `back-end/apps/api/src/auth/auth.service.ts`:
- Add an import for Node crypto (`randomInt`) near existing imports.
- Update `generatePassword()` so each character index is produced with `randomInt(26)` instead of `Math.floor(Math.random() * 26)`.

No other methods or signatures need to change.

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._

### DIFF
--- a/automation/pages/OrganizationPage.ts
+++ b/automation/pages/OrganizationPage.ts
@@ -21,6 +21,7 @@ import {
   verifyOrganizationExists,
 } from '../utils/db/databaseQueries.js';
 import * as fs from 'node:fs';
+import { randomInt } from 'node:crypto';
 import { argonHash } from '../utils/crypto/crypto.js';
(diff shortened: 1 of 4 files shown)


svienot (approved):
LGTM

PR #3099, opened 2026-07-10
fix: notification badge inflated count

### LINKED ISSUE #3098: fix: notification badge inflated when transaction expires
## Bug

When a transaction expires (moves from *Waiting for Signatures* → *Expired*), the in-app notification badge increments by 2 or 3 instead of 1. This is most visible for the transaction creator, whose badge starts at 0, making the inflation unmistakable.

## Root cause

`listenForUpdates()` in `storeNotifications.ts` registers socket.io listeners for `notifications:new` and `notifications:indicators:delete` every time `ws.setup()` is called, but never removes the previous ones. The disposer functions returned by `ws.on()` were being discarded.

`ws.setup()` is called 3 times during a normal login/startup sequence:

1. `user.login()` → `selectOrganization(null)` → `afterOrganizationSelection()` → `setupStores()` (no orgs loaded yet, no listeners added)
2. `selectDefaultOrganization()` → `selectOrganization(org)` → `afterOrganizationSelection()` → `setupStores()` (listeners registered)
3. `setupStores()` called directly in `GlobalAppProcesses.vue` / `EmailLoginForm.vue` (listeners stacked again)

Calls 2 and 3 both run after orgs are populated, so two copies of each listener end up on the same socket. One backend event fires both, inflating the badge count.

## Fix

Save the disposers returned by `ws.on()` and call them at the top of `listenForUpdates()` before re-registering, making the function idempotent regardless of how many times `ws.setup()` fires.

jbair06:
**Description**:

When a transaction expires, the notification badge incremented by 2–3 instead of 1. This was most visible for the transaction creator, whose badge starts at 0 before the expiry event, making the inflation unmistakable.

`listenForUpdates()` in `storeNotifications.ts` registers socket.io listeners for `notifications:new` and `notifications:indicators:delete` on every `ws.setup()` call without removing the previous ones — the disposer functions returned by `ws.on()` were being discarded. `ws.setup()` fires 3 times during a normal login/startup sequence, so two copies of each listener ended up on the same socket. One backend event fired both, doubling (or tripling) the badge increment.

Fix: save the disposers returned by `ws.on()` and call them at the top of `listenForUpdates()` before re-registering, making the function idempotent regardless of how many times `ws.setup()` fires. The underlying socket connection is unaffected — no open/close occurs between the redundant setup calls, so there is no message-loss window.

**Related issue(s)**:

Fixes #3098

### DIFF
--- a/front-end/src/renderer/stores/storeNotifications.ts
+++ b/front-end/src/renderer/stores/storeNotifications.ts
@@ -109,6 +109,7 @@ const useNotificationsStore = defineStore('notifications', () => {
   });
 
   let notificationsQueue = Promise.resolve();
+  let notificationListenerDisposers: (() => void)[] = [];
 
   /** Preferences **/
   async function fetchPreferences() {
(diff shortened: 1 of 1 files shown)


svienot (approved):
LGTM

PR #3112, opened 2026-07-16
chore: update pnpm to version 11.15.1

svienot:
**Description**:

- Upgrade version of  pnpm to 11.15.1 (in package.json and Dockerfiles)
- Move overrides section from package.json to pnpm-workspace.yaml
- Remove .npmrc and move its content to pnpm-workspace.yaml
- Add trustLockfile: true to preserve previous pnpm behavior
- Add override for postcss to avoid last version too recent to pass the cool-down check

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,2 +1,0 @@
-virtual-store-dir-max-length=60
-shamefully-hoist=true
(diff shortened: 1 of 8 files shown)


ericleponner (approved):
LGTM

Issue #2936, opened 2026-06-07
Encrypt JWT token at rest and reduce token lifetime


jbair06:
**Description:**
The Electron client stores session tokens in a way that does not meet the same protection standard applied to other credentials in the local database. Token lifetime is also excessive. Both issues need to be addressed together.

**Acceptance Tests:**
- Session tokens are encrypted at rest using the same mechanism that protects other credentials in the local database
- Token lifetime is significantly reduced and a refresh token mechanism is implemented for renewal
- An "active sessions" view and "log out everywhere" action allow administrators to invalidate tokens after a suspected device compromise

**Notes:**
Any future "remember me" feature must not regress token storage to a weaker protection standard.

PR #3124, opened 2026-07-20
chore:  now passing Organization in place of serverUrl

ericleponner:
**Description**:

Changes below update functions that expect `serverUrl` : they now expect an `Organization` instance.This gives these functions access to `org.serverUrl` and also `org.id` (which enables to retrieve JWT token).

**Related issue(s)**:

Contributes to #2936

**Notes for reviewer**:

**Checklist**


ericleponner:
We'll do this refactoring later. Closing…

PR #3137, opened 2026-07-22
fix: Fix and harden transaction comment creation

### LINKED ISSUE #2933: Fix broken comment creation endpoint and add access control gate
**Description:**
`POST /transactions/:transactionId/comments` has two compounding problems: the handler crashes on every request due to a broken relation assignment in the service, and there is no check that the requester has any relationship to the target transaction. A TODO comment in the controller explicitly flags the missing gate. Neither issue can be addressed in isolation — fixing the storage path without adding the access check silently reopens an unauthorized-comment capability.

**Acceptance Tests:**
- Authenticated user with no relationship to a transaction receives 403 when attempting to post a comment
- Authenticated user with a valid relationship (creator, signer, approver) can post a comment successfully
- `CreateCommentDto.message` enforces a maximum length (≤ 2,000 characters); requests exceeding it receive 400
- The service no longer throws `TypeError` on valid requests
- Response shape is controlled via a `Serialize` DTO and does not leak unintended relation fields

**Notes:**
Fix the relation assignment in `comments.service.ts` (the broken `comment['transaction'].id` pattern) using an explicit relation object. Gate the handler with the same `verifyAccess` check used by the transaction-view endpoint. Add per-user rate limiting. The entity column and DTO constraint should be updated together to keep them in sync.

svienot:
**Description**:

- Fix the `POST /transactions/:transactionId/comments` endpoint
- Add database constraint for the length of the comment message (2000 char max).
- Apply `UserThrottlerGuard` to the comment creation endpoint
- Note that the check that the requester has any relationship to the target transaction has already been implemented as part of #3020 (`TransactionAccessGuard` is applied to all the routes of the `CommentsController`

**Related issue(s)**:

Fixes #2933 

 **/!\ Needs to be merged with main once #3157 has been integrated**

### DIFF
--- a/back-end/apps/api/src/transactions/comments/comments.controller.spec.ts
+++ b/back-end/apps/api/src/transactions/comments/comments.controller.spec.ts
@@ -7,7 +7,7 @@ import { User } from '@entities';
 
 import { TransactionCommentDto } from '../dto';
 
-import { VerifiedUserGuard } from '../../guards';
+import { UserThrottlerGuard, VerifiedUserGuard } from '../../guards';
 import { TransactionAccessGuard } from '../../guards/transaction-access.guard';
 
 import { CommentsController } from './comments.controller';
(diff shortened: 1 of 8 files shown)


ericleponner (approved):
LGTM

PR #3158, opened 2026-07-23
chore: moving to typescript 5.8.3 and vue to 3.5.40

ericleponner:
**Description**:

Change below move `typescript` to 5.8.3, `vue` to 3.5.40, `vite` to 7.3.6, `pinia` to 4.0.2.
They also update some dependent components.

This should make #3149 very close to pass.


svienot (approved):
LGTM

PR #3162, opened 2026-07-24
chore: moving to latest eslint-plugin-vue.

ericleponner:
**Description**:

Changes below upgrade `eslint-plugin-vue` and `@vue/eslint-config-typescript` to their latest versions.

Those new tools trigger a lot of ESLint errors/warnings (~600):
- most obvious/safe are fixed by changes below
- some of them are hidden by disabling some eslint rules
   - many concerned test code (mock related)
   - those rules will have to be re-enabled and code fixed later


svienot (approved):
LGTM

PR #3164, opened 2026-07-24
chore: migrating various dependencies

ericleponner:
**Description**:

Changes below migrate various front-end / back-end  dependencies to their best releases (ie which enables to compile and run TT).


svienot (approved):
LGTM

PR #3106, opened 2026-07-15
chore: Auth token is no longer duplicated in sessionStorage.

ericleponner:
**Description**:

Changes below remove logic which duplicates `auth token` in `sessionStorage`.
In current code, `auth token` is maintained in two places:
1) local db (`OrganizationCredentials.jwtToken`)
2) browser `sessionStorage`

Storing token in `sessionStorage` is not necessary because it is available from `userStore` everywhere.

**Related issue(s)**:

Contributes to #2936

**Notes for reviewer**:

```
M   front-end/src/renderer/services/organization/reconnect.ts
    # Call to getAuthTokenFromSessionStorage() replaced by storeUser.getJwtToken().
    # Call to toggleAuthTokenInSessionStorage() removed

M   front-end/src/renderer/stores/storeUser.ts
    # Call to setSessionStorageTokens() removed

M   front-end/src/renderer/utils/axios.ts
M   front-end/src/renderer/stores/storeWebsocketConnection.ts
    # Call to getAuthTokenFromSessionStorage() replaced by storeUser.organizations.find() + storeUser.getJwtToken()

M   front-end/src/renderer/pages/Migrate/components/PerformSetup.vue
M   front-end/src/renderer/pages/OrganizationLogin/OrganizationLogin.vue
M   front-end/src/renderer/pages/Settings/components/OrganizationsTab.vue
M   front-end/src/renderer/pages/Settings/components/ProfileTab.vue
M   front-end/src/renderer/services/organization/disconnect.ts
    # Removed call to toggleAuthTokenInSessionStorage()

M   front-end/src/renderer/utils/userStoreHelpers.ts
    # Removed now obsolete SESSION_STORAGE_AUTH_TOKEN_PREFIX,
    # setSessionStorageTokens(), toggleAuthTokenInSessionStorage()
    # and getAuthTokenFromSessionStorage().

M   front-end/src/tests/renderer/pages/OrganizationLogin/OrganizationLogin.spec.ts
    # Updated tests related to toggleAuthTokenInSessionStorage()

M   front-end/src/tests/renderer/services/organization/reconnect.spec.ts
    # Updated tests related to getAuthTokenFromSessionStorage()
```

**Checklist**

- [ ] Documented (Code comments, README, etc.)
- [x] Tested (unit, integration, etc.)

PR #3169, opened 2026-07-27
fix: fixing regression in migration wizard

ericleponner:
**Description**:

Changes below fix a regression introduced in #3058 (and revealed by #3106) and adds an optimization:
- `PerformSetup.setupOrganization()` now makes sure to refresh user store before calling`changePassword()` so this one can retrieve JWT token when connecting to backend
- `SetupOrganization` now propagates JWT token to `PerformSetup` which avoids an extra call to `login()` and some possible `429` errors.

**Notes for reviewer**:
```
M   front-end/src/renderer/pages/Migrate/components/PerformSetup.vue
    # setupOrganization() now makes sure to refresh user/organization stores before calling changePassword().
```

**Checklist**

### DIFF
--- a/front-end/src/renderer/pages/Migrate/Migrate.vue
+++ b/front-end/src/renderer/pages/Migrate/Migrate.vue
@@ -56,6 +56,7 @@ const recoveryPhrase: Ref<RecoveryPhrase | null> = ref(null);
 const recoveryPhrasePassword = ref<string | null>(null);
 const personalUser = ref<PersonalUser | null>(null);
 const organizationSetup = ref<ModelValue | null>(null);
+const jwtToken = ref<string | null>(null);
 
 const userInitialized = ref(false);
 const keysImported = ref(0);
@@ -116,8 +117,9 @@ const handleSetPersonalUser = async (value: PersonalUser) => {
   step.value = 'organization';
 };
 
-const handleSetOrganizationSetup = async (value: ModelValue | null) => {
+const handleSetOrganizationSetup = async (value: ModelValue | null, token: string | null) => {
   organizationSetup.value = value;
+  jwtToken.value = token;
   await initializeUserStore();
   userInitialized.value = true;
   // Write the skip claim now that the org (if any) is selected, using the correct claim key.
@@ -229,6 +231,7 @@ const initializeUserStore = async () => {
           <PerformSetup
             :personal-user="personalUser!"
             :organization-setup="organizationSetup"
+            :jwt-token="jwtToken!"
             :recovery-phrase="recoveryPhrase"
             :recovery-phrase-password="recoveryPhrasePassword"
             :selected-keys="selectedKeysToRecover"
--- a/front-end/src/renderer/pages/Migrate/Migrate.vue
+++ b/front-end/src/renderer/pages/Migrate/Migrate.vue
@@ -236,7 +236,6 @@ const initializeUserStore = async () => {
             :recovery-phrase-password="recoveryPhrasePassword"
             :selected-keys="selectedKeysToRecover"
             @didPerformSetup="didPerformSetup"
-            @didCancelSetup="handleStopMigration"
           />
         </template>
 
--- a/front-end/src/renderer/pages/Migrate/Migrate.vue
+++ b/front-end/src/renderer/pages/Migrate/Migrate.vue
@@ -227,7 +227,7 @@ const initializeUserStore = async () => {
         </template>
 
         <!-- Perform Migration Step -->
-        <template v-if="stepIs('performSetup') && personalUser !== null && jwtToken !== null">
+        <template v-if="stepIs('performSetup') && personalUser !== null">
           <PerformSetup
             :personal-user="personalUser"
             :organization-setup="organizationSetup"
--- a/front-end/src/renderer/pages/Migrate/Migrate.vue
+++ b/front-end/src/renderer/pages/Migrate/Migrate.vue
@@ -227,15 +227,16 @@ const initializeUserStore = async () => {
         </template>
 
         <!-- Perform Migration Step -->
-        <template v-if="stepIs('performSetup')">
+        <template v-if="stepIs('performSetup') && personalUser !== null && jwtToken !== null">
           <PerformSetup
-            :personal-user="personalUser!"
+            :personal-user="personalUser"
             :organization-setup="organizationSetup"
-            :jwt-token="jwtToken!"
+            :jwt-token="jwtToken"
             :recovery-phrase="recoveryPhrase"
             :recovery-phrase-password="recoveryPhrasePassword"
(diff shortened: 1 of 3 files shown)


front-end/src/renderer/pages/Migrate/components/PerformSetup.vue
```diff
@@ -54,37 +63,45 @@ const concludeSetup = async (importedKeyCount: number, error: unknown) => {
...
-  const { jwtToken } = await login(
-    setup.organizationURL,
+  await addOrganizationCredentials(
     email,
     setup.temporaryOrganizationPassword,
+    organizationId,
+    props.personalUser.personalId,
+    props.jwtToken,
+    null,
   );
```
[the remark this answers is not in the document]
ericleponner: I prefer to keep jwtToken not nullable.

front-end/src/renderer/pages/Migrate/Migrate.vue
```diff
@@ -229,6 +231,7 @@ const initializeUserStore = async () => {
           <PerformSetup
             :personal-user="personalUser!"
             :organization-setup="organizationSetup"
+            :jwt-token="jwtToken!"
             :recovery-phrase="recoveryPhrase"
```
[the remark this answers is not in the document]
ericleponner: I prefer to keep jwtToken not nullable.

svienot (dismissed):
LGTM

front-end/src/renderer/pages/Migrate/Migrate.vue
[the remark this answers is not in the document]
jbair06: When the org setup step is skipped, it appears to hang showing only 'Setup'. `jwtToken` should be nullable, allowing the user to 'skip' the org setup.

front-end/src/renderer/pages/Migrate/components/PerformSetup.vue
```diff
@@ -54,37 +64,50 @@ const concludeSetup = async (importedKeyCount: number, error: unknown) => {
...
+  await updateOrganizationCredentials(
     organizationId,
     props.personalUser.personalId,
-    jwtToken,
-    props.personalUser.password,
-    true,
+    undefined,
+    undefined,
+    undefined,
+    encryptedNewPassword,
```
jbair06: The 6th argument should be the password used to encrypt the new password. I don't believe the already encrypted password should be passed in here.

front-end/src/renderer/pages/Migrate/components/PerformSetup.vue
```diff
@@ -54,37 +64,50 @@ const concludeSetup = async (importedKeyCount: number, error: unknown) => {
...
+  // 1) Encrypt new password
+  const personalPassword = await getPasswordAsync({
+    subHeading: 'Enter your application password to encrypt your organization credentials',
+  });
+  if (personalPassword === false) {
+    // User has refused to enter his application password => setup is cancelled
+    emit('didCancelSetup');
+    return;
+  }
+  const encryptedNewPassword = await encryptOrganizationPassword(
```
jbair06: `addOrganization` and `updateOrganization` handles the encryption of the new password. Does it need to be done outside of these functions?

front-end/src/renderer/pages/Migrate/components/PerformSetup.vue
```diff
@@ -54,37 +64,50 @@ const concludeSetup = async (importedKeyCount: number, error: unknown) => {
...
 const setupOrganization = async (setup: ModelValue) => {
-  // 1) Add organization
+  // 1) Encrypt new password
+  const personalPassword = await getPasswordAsync({
+    subHeading: 'Enter your application password to encrypt your organization credentials',
+  });
+  if (personalPassword === false) {
+    // User has refused to enter his application password => setup is cancelled
+    emit('didCancelSetup');
+    return;
```
jbair06: This return goes back to `onMounted` which would continue with the setup. Should `didCancelSetup` also cancel the rest of what happens in `onMounted`?

front-end/src/renderer/pages/Migrate/components/PerformSetup.vue
```diff
@@ -54,37 +64,50 @@ const concludeSetup = async (importedKeyCount: number, error: unknown) => {
...
-  const email = setup.organizationEmail!; // Checked by SetupOrganization.checkLoginInOrganization()
-  const { jwtToken } = await login(
-    setup.organizationURL,
+  await addOrganizationCredentials(
     email,
     setup.temporaryOrganizationPassword,
+    organizationId,
+    props.personalUser.personalId,
+    props.jwtToken,
+    null,
```
jbair06: This call used to be made with `updateIfExists` to false, to prevent any issues if the org already exists. This should _never_ actually happen, but should we add that `false` back in just in case?

jbair06 (changes requested):

front-end/src/renderer/pages/Migrate/Migrate.vue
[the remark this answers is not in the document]
ericleponner: Indeed `jwtToken` should be nullable.
Fixed : `PerformSetup.onMounted()` now runs `setupOrganization()` when `props.organizationSetup !== null` AND `props.jwtToken !== null`.

front-end/src/renderer/pages/Migrate/components/PerformSetup.vue
ericleponner: In fact, it used to be called with [updateIfExists == true](https://github.com/hashgraph/hedera-transaction-tool/blob/b16b06ba9f36f2063a381bc0170da0704250b11e/front-end/src/renderer/pages/Migrate/components/SetupOrganization.vue).

We can call it with `false`. But that means `SetupOrganization` view should ideally check if org exists and prevent to go further in the migration process if it's the case.

Let's go with `updateIfExists = false` for now…

front-end/src/renderer/pages/Migrate/components/PerformSetup.vue
ericleponner: If my understanding is correct, `setupOrganization()` does not need to call `getPasswordAsync()`: `personalPassword` is available in `props.personalUser.password`.
Removing call to `getPasswordAsync()` and call to `emit('didCancelSetup')`.

front-end/src/renderer/pages/Migrate/components/PerformSetup.vue
ericleponner: Ok, I get it now.
There is no need for calling `getPasswordAsync()` : `props.personalUser.password` already contains the value.

So call to `addOrganizationCredentials()` can be:

Image

And call to `updateOrganizationCredentials()`:

Image

jbair06 (approved):

PR #3172, opened 2026-07-29
chore: bump  prisma related dependencies

svienot:
**Description**:

- Bump the following dependencies (See initial dependabot PR for full details: #3148):
  - @prisma/adapter-better-sqlite3 from 7.8.0 to 7.9.1
  - @prisma/client from 7.8.0 to 7.9.1
  - prisma from 7.8.0 to 7.9.1
- Update overrides
- Fix latent bug in automation utility function `reloadSession` which caused `transactionAccountDeleteTests.test.ts` to fail. `reloadSession` restarts the tool with `preserveLocalState: true`, so the key initially imported in the beforeEach remains in SQLite. `reloadSession` was calling `setupTransactionEnvironment` which caused the same key import to occur again.


ericleponner (approved):
LGTM

PR #3182, opened 2026-08-01
fix: provide user feedback from failed signing process

### LINKED ISSUE #2865: Missing key during signing from "Ready to Sign" tab provides no user feedback
**Description:**  
When attempting to sign a transaction from the "Ready to Sign" tab with a missing key, the user receives no feedback and it appears as if no action occurred. This differs from signing within the transaction details view, where appropriate feedback is shown.

**Expected behavior:**  
- Users receive clear feedback when attempting to sign without a required key from the "Ready to Sign" tab.  
- Feedback is consistent with the messaging shown in the transaction details view.

**Acceptance criteria:**  
- Attempting to sign without a required key from the "Ready to Sign" tab displays an appropriate error or warning message.  
- Feedback messaging matches or aligns with the transaction details signing experience.  
- No silent failures occur during signing attempts.

jbair06:
**Description**:
When signing fails due to a missing key from the Ready to Sign tab or group details page, the error modal was being destroyed before it could render — because callbacks always triggered a refresh or navigation regardless of outcome. Fixed by threading the success/failure result through the callback chain so refreshes and navigation only happen when signing actually succeeds.

**Related issue(s)**:

Fixes #2865 

**Checklist**

### DIFF
--- a/front-end/src/renderer/pages/TransactionGroupDetails/SignAllController.vue
+++ b/front-end/src/renderer/pages/TransactionGroupDetails/SignAllController.vue
@@ -37,13 +37,13 @@ const progressText = ref<string>('');
 
 /* Handlers */
 const handleSignAll = async (userPersonalPassword: string | null): Promise<ActionReport | null> => {
-  let result: ActionReport | null;
+  let result: ActionReport | null = null;
   if (props.groupOrId !== null) {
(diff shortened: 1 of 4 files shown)


ericleponner (approved):
LGTM

